PR #1, opened 2022-06-02
Setting up build.gradle files and dependency files

ramsessanchez:
Hi vincent, I have set up the build.gradle files, these are similar to the ones we have set up for the current v1 and beta sdk's. I have left some spaces with the placeholder [NAME], let me know if you think the naming conventions are correct in these spaces on just one of the build.gradle files and I'll be sure to replicate the changes for all the separate components. 
Also, I left in the publishing tasks in there, along with the snapshot ones, I believe we don't need them for the time being but have left them in to facilitate the work later on. Let me know if you think otherwise. Thanks!

### DIFF
--- a/components/abstractions/build.gradle
+++ b/components/abstractions/build.gradle
@@ -1,36 +1,219 @@
 plugins {
-    // Apply the java-library plugin for API and implementation separation.
+    // Apply the java-library plugin to add support for Java Library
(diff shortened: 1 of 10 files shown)


ramsessanchez:
I will be adding the settings and some script files that this build file uses on a separate pr to make review easier.

PR #2, opened 2022-06-02
Properties/Scripts

ramsessanchez:
Adding the Properties and Scripts files.
Added placeholder for patch version, do we begin each of these at version 0.0.1 since these are all being added to a new repo-housing space? Or do we keep the values we have set on the original Kiota repo? 
Not included is the update minor-version script, this will be added with the github workflows once the structure for them is decided.

### DIFF
--- a/components/abstractions/gradle.properties
+++ b/components/abstractions/gradle.properties
@@ -1,0 +1,41 @@
+# IDE users:
+# Settings specified in this file will override any Gradle settings
+# configured through the IDE.
+
+# For more details on how to configure your build environment visit
+# http://www.gradle.org/docs/current/userguide/build_environment.html
+
+# Specifies the JVM arguments used for the daemon process.
+# The setting is particularly useful for tweaking memory settings.
+# Default value: -Xmx10248m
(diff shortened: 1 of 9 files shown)


components/abstractions/gradle.properties
```diff
@@ -0,0 +1,41 @@
...
+# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
+# org.gradle.parallel=true
+
+# The size of the library demands a large amount of RAM to build. Increase as necessary if you get GC errors
+## linux requires 10G, OSX requires 11G
+org.gradle.jvmargs=-Xmx2g
+org.gradle.parallel=true
+org.gradle.caching=true
+
+mavenGroupId         = com.microsoft.graph
```
baywet, line 24:
```suggestion
mavenGroupId         = com.microsoft.kiota
```

components/abstractions/gradle.properties
```diff
@@ -0,0 +1,41 @@
...
+mavenArtifactId      = microsoft-graph-kiota-java-[NAME]
```
baywet, line 25:
```suggestion
mavenArtifactId      = microsoft-kiota-[NAME]
```

baywet (changes requested):

PR #25, opened 2022-08-12
Verifies URL formats

andrueastman:
This PR is related to https://github.com/microsoft/kiota-abstractions-dotnet/pull/30

It verifies URL formats in the case of boolean and Date path parameters are added to the path parameters collection by adding tests to verify the behaviour.

Changes include
- Adds tests to verify DateTime and DateTimeOffsets default to ISO 8601. 
- Adds check to throw IllegalStateException when the baseUrl path parameter is not set.

### DIFF
--- a/components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
@@ -49,7 +49,7 @@ public class RequestInformation {
         } else {
             Objects.requireNonNull(urlTemplate);
             Objects.requireNonNull(queryParameters);
-            if(urlTemplate.toLowerCase().contains("{+baseurl}") && !pathParameters.containsKey("baseurl"))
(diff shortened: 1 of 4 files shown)


components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
```diff
@@ -48,6 +49,9 @@ public URI getUri() throws URISyntaxException {
         } else {
             Objects.requireNonNull(urlTemplate);
             Objects.requireNonNull(queryParameters);
+            if(urlTemplate.toLowerCase().contains("{+baseurl}") && !pathParameters.containsKey("baseurl"))
```
baywet, line 52:
```suggestion
            if(!pathParameters.containsKey("baseurl") && urlTemplate.toLowerCase().contains("{+baseurl}"))
```
to avoid checking the string when we already have an entry

baywet (changes requested):
thanks for the changes. Would you mind adding a changelog entry as well please?

baywet (approved):

PR #29, opened 2022-08-29
- adds support for composed types deserialization in abstractions

baywet:
related https://github.com/microsoft/kiota/pull/1603


ramsessanchez (approved):

ramsessanchez:
@baywet OkHttp and Serialization checks wont pass since they take a dependency on the changes to abstractions that you've made. If it all builds locally I suggest we allow this to pass.

PR #37, opened 2022-09-07
Update dependencies.gradle

ramsessanchez:
Bump abstractions snapshot dependency in Auth Component.


baywet (changes requested):
While we're at it we should probably make abstractions an Api dependency for all the implementation packages

baywet (approved):

PR #41, opened 2022-09-08
feature/status codes

baywet:
related https://github.com/microsoft/kiota/issues/1699


ramsessanchez:
@baywet can confirm that this doesn't fix the issue with the call to client.me().drive(), however, that seems unrelated and is likely to be a service issue. Merging this.

ramsessanchez (approved):

PR #42, opened 2022-09-14
source and target compatibility set to java version 8

ramsessanchez:


components/abstractions/build.gradle
```diff
@@ -150,8 +150,8 @@ def fixAscNames = { name ->
 
 compileJava {
   options.compilerArgs << "-parameters"
-  sourceCompatibility = '17'
-  targetCompatibility = '17'
+  sourceCompatibility = '11'
+  targetCompatibility = '11'
```
baywet, line 154:
```suggestion
  sourceCompatibility = '17'
  targetCompatibility = '8'
```

components/authentication/azure/build.gradle
```diff
@@ -143,8 +143,8 @@ def fixAscNames = { name ->
 
 compileJava {
   options.compilerArgs << "-parameters"
-  sourceCompatibility = '17'
-  targetCompatibility = '17'
+  sourceCompatibility = '11'
+  targetCompatibility = '11'
```
baywet, line 147:
```suggestion
  sourceCompatibility = '17'
  targetCompatibility = '8'
```

components/http/okHttp/build.gradle
```diff
@@ -144,8 +144,8 @@ def fixAscNames = { name ->
 
 compileJava {
   options.compilerArgs << "-parameters"
-  sourceCompatibility = '17'
-  targetCompatibility = '17'
+  sourceCompatibility = '11'
+  targetCompatibility = '11'
```
baywet, line 148:
```suggestion
  sourceCompatibility = '17'
  targetCompatibility = '8'
```

components/serialization/text/build.gradle
```diff
@@ -143,8 +143,8 @@ def fixAscNames = { name ->
 
 compileJava {
   options.compilerArgs << "-parameters"
-  sourceCompatibility = '17'
-  targetCompatibility = '17'
+  sourceCompatibility = '11'
+  targetCompatibility = '11'
```
baywet, line 147:
```suggestion
  sourceCompatibility = '17'
  targetCompatibility = '8'
```

baywet (changes requested):

baywet (dismissed):

ramsessanchez:
@baywet I believe this is ready to go. However, I think we should merge the two yml file updates in order to test their functionality. 
I think this #45  should be merged first and then #46. This would allow us to test functionality. 
Also, we expect all components aside from abstractions to fail since they all rely on updates to the abstractions component before they are able to build successfully. Considering this is a common occurrence which has prompted our use of manual releases, I was wondering how you felt about having any changes made to abstractions live in their own PR's? An alternative I think would also be a specialized 'dev-abstractions' branch where we release from so that when changes are later made to other components they are able to pass the build check and subsequently release automatically upon merging.

ramsessanchez:
@baywet This is ready. I upgraded the abstractions to 0.0.4 so the releases for the other components will fail as they are using 0.0.3 and this will cause compatibility issues. I will make a separate pr to update the dependency version on these which will allow us to further test the build and release pipelines.

baywet (approved):

baywet:
@ramsessanchez check paths is still failing https://github.com/microsoft/kiota-java/actions/runs/3062809845/jobs/4944193457

PR #43, opened 2022-09-14
feature/devcontainer

baywet:
the dev container is impacted by this https://github.com/microsoft/vscode-dev-containers/issues/1628

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,23 @@
+# [Choice] Java version (use -bullseye variants on local arm64/Apple Silicon): 11, 17, 11-bullseye, 17-bullseye, 11-buster, 17-buster
+ARG VARIANT=11-bullseye
+FROM mcr.microsoft.com/vscode/devcontainers/java:${VARIANT}
+
+# [Option] Install Maven
+ARG INSTALL_MAVEN="false"
+ARG MAVEN_VERSION=""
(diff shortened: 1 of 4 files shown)


ramsessanchez:
Quick question, I didn't notice a devContainer folder or Dockerfile in our current core/client libraries so I was just hoping you could expand on the purpose of adding it here? Thanks!

baywet:
The idea is to be able to use codespaces, or run from a container locally, instead of having to install java & code on your host OS. You currently only work on Java, kiota, and typewriter (dotnet), so I don't expect that to make a big difference to you. But in my case I'm working in dotnet, java, go, typescript, (ruby, php, python, swift...) and getting tired of updating tooling every other day on my machine 😅

baywet:
as per the vs code files, they are just here to provide a better experience when using vs code, regardless of where vs code is running.

ramsessanchez:
> The idea is to be able to use codespaces, or run from a container locally, instead of having to install java & code on your host OS. You currently only work on Java, kiota, and typewriter (dotnet), so I don't expect that to make a big difference to you. But in my case I'm working in dotnet, java, go, typescript, (ruby, php, python, swift...) and getting tired of updating tooling every other day on my machine 😅

Shoulda stuck to program management. Thanks for the clarification!

ramsessanchez (approved):

PR #45, opened 2022-09-14
Update gradleBuild-trigger.yml

ramsessanchez:
Allow for building on workflow_dispatch as well as running a test build when a specific component is changed
Fixes the issue where every component would be built on a PR even if no changes were made to that component.


baywet (dismissed):

ramsessanchez:
@baywet, we can force this merge

baywet (approved):

PR #67, opened 2022-10-11
- bumps all packages version following android compatibility

baywet:
Signed-off-by: Vincent Biret <vibiret@microsoft.com>


ramsessanchez:
@baywet This should be reverted or the readme's should be updated. I had bumped the versions in the android/compatibility branch so this leads to a an extra bump.

ramsessanchez:
@baywet I'll bump the readme's since were implemented these versions in core.

PR #68, opened 2022-10-11
Changelogs

ramsessanchez:
Versioning for changelogs updated to match current versions
CompletableFutures adjusted for  level 26 in the http component. Passes linting.


baywet (approved):
Thanks, I forgot we already had bumped

Issue #35, opened 2022-09-01
fix worklfows so the library releases works without manual intervention


baywet:
currently the release doesn't tick automatically because of the file filter steps, needs to be corrected

PR #118, opened 2023-01-04
Re-structure using gradle multi-project

### LINKED ISSUE #102: consider breaking up this repo in sub repository
considering the issues we've had with the CI complexity #35 and the fact that SonarCloud doesn't support monorepos #101, we should consider breaking this repo appart in sub repos like the other languages.

andreaTP:
Proposal to fix #102 

This is a first step to go full into using multi-project builds, which are pretty common in the Java ecosystem.
I'm using IntelliJ as my favourite IDE for Java projects and with this setup you can directly import the root folder once and have all the sub-projects imported as well.

I would need some coordination, guidance and support to test out the release steps, hence the Draft state of this PR.
Please note that the CI and the build definitions can be further simplified, but we need to take a couple of decisions before getting there.


.github/workflows/api-level-lint.yml
```diff
@@ -10,146 +10,23 @@ on:
...
-  lint-api-level-abstractions:
+  lint-api-level:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-java@v3
         with:
           distribution: 'adopt'
-          java-version: 18
+          java-version: 17
```
andreaTP, line 20: I included this change of the Java version since 18 is not an LTS:
https://www.oracle.com/java/technologies/java-se-support-roadmap.html

.github/workflows/gradleBuild-component.yml
```diff
@@ -13,31 +13,27 @@ on:
...
         uses: actions/upload-artifact@v3
         with:
           name: Unit Tests
           path: |
-              build/reports/tests/test/**
-              build-test-results/**
+              components/**/build/reports/tests/test/**
+              components/**build-test-results/**
+      # TODO: not sure what to do with the next step
       - name: Upload a Build Artifact
```
baywet, line 37: what's collected below probably needs to be updated since we're not setting the working directory anymore

.github/workflows/release-component.yml
```diff
@@ -55,7 +52,5 @@ jobs:
         env: 
           ENCODED_VALUE: ${{ secrets.SECRING_GPG }}
           OUTPUT_PATH: '.\secring.gpg'
-      - name: Grant execute permission for gradlew
-        run: chmod +x gradlew
       - name: Publish
-        run: ./gradlew $PREVIEW_TASK
+        run: ./gradlew $PREVIEW_TASK -p ${{ inputs.workingDirectory }}
```
baywet, line 56: if I'm following properly this is how the release targets a specific component now?

components/serialization/text/CHANGELOG.md
```diff
@@ -1,26 +0,0 @@
-# Changelog
-
-All notable changes to this project will be documented in this file.
-
-The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
```
baywet, line 5: removing the multiple changelogs won't work unless we agree on aligning the versions across components. Can you please revert that aspect until we reach a conclusion on #119?

settings.gradle
```diff
@@ -0,0 +1,19 @@
+rootProject.name = 'kiota-java'
+
+include ':components:abstractions'
```
baywet, line 3: out of curiosity why not put this block in the else instead of logging a warning?

baywet (changes requested):
Thank you so much for the contribution! It'll make development cycles easier. I left a couple of comments, don't hesitate if you have follow up on them

.github/workflows/gradleBuild-component.yml
```diff
@@ -13,31 +13,27 @@ on:
...
         if: ${{ always() }}
         uses: actions/upload-artifact@v3
         with:
           name: Unit Tests
           path: |
-              build/reports/tests/test/**
-              build-test-results/**
+              components/**/build/reports/tests/test/**
+              components/**build-test-results/**
+      # TODO: not sure what to do with the next step
```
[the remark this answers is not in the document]
andreaTP: As far as I understand it, uploading the test results is correctly performed here:
https://github.com/microsoft/kiota-java/blob/2960c9ee3ef104cdba0baa21d8108c18f73f3df6/.github/workflows/api-level-lint.yml#L35

I'm missing the reason to upload gradle-related things and gradle-generated files (they should be 100% reproducible).

Would you agree on uploading only the test results?

.github/workflows/gradleBuild-component.yml
andreaTP: As commented above, I'm proposing to change this upload to only upload test results.

.github/workflows/release-component.yml
andreaTP: Correct, as mentioned in the PR this is not the final step of the refactoring, it should be possible to release everything with a single command (almost 🙂 ) from the root after those changes.

components/serialization/text/CHANGELOG.md
andreaTP: Ok, I revert this change, I felt the freedom of removing them as I noticed that most are outdated in different ways, but I agree on reverting 👍

settings.gradle
andreaTP: if I understood correctly the current setup those modules are always available, but the android modules should be compiled(in addition to the standard) only if the relevant SDK is installed.

Putting those in the `else` branch disables the modules from being built/tested/compiled when you have the SDK installed, is it the expected behavior?

Just to make it clear, I have almost 0 experience with Android development, and I'm refactoring with plain Java usage in mind, feel free to challenge and eventually correct my decisions and assumptions if you feel like 👍

.github/workflows/gradleBuild-component.yml
andreaTP: Looking again at this file and I think that the original intention was to upload the built artifacts, hence proposing this change:
https://github.com/microsoft/kiota-java/pull/118/commits/459c804b346e73ec7ae7984a5da09275e5bd0790

Gradle-related files should be irrelevant.

andreaTP:
@baywet ready for next round 🙂

.github/workflows/gradleBuild-component.yml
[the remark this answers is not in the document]
baywet: you are correct in the sense that we're also uploading these results which are to ensure we're not using any API from a level above the one we've set.
and we also want the test results to be uploaded. The reasoning is that not everyone is always on their computer with the repository cloned and the tooling installed, and having access to the results without those pre-requisites (e.g. when talking about a failing PR on a meeting) is helpful.

settings.gradle
baywet: right, sorry I was multi-tasking, and I'm not great at doing so.
the idea of the android projects is just to check we're not using APIs from API levels above the one we're targeting (android only has a partial implementation of the java 8 spec and they release additional implementations with newer API levels).
during our CI we run both (with different workflows), but during publish only the "pure java" project is being used.

With that context I think we can simply remove the reference to the android projects here, and add a new root project in the android directory, makes sense?

.github/workflows/gradleBuild-component.yml
baywet: as mentioned in the other comment, we upload 3 things:
- android api level linting report (separate workflow)
- build artifacts
- unit test results
- (some of the sources, but I don't think it was any use and you just removed that part rightfully)

baywet:
@andreaTP thank you for your responsiveness on the feedback. Here is another round of answers :)

andreaTP:
Thanks for all the feedback! I'll go through everything tomorrow morning (for me 😊).

.github/workflows/gradleBuild-component.yml
andreaTP: Changed to upload all of the build artifacts: `./**/build/*`, build results are explicitly and separately uploaded in the step above.

settings.gradle
andreaTP: @baywet thanks a lot for this explanation, very much needed!
I think that this commit correctly serves the purpose: https://github.com/microsoft/kiota-java/pull/118/commits/3e2e9962d573a1259487cb5c6d2f9fbff0992247

We can probably simplify the android setup, but let's keep it for another PR.

andreaTP:
Tested some parts of the CI on my fork and it looks better now:
https://github.com/andreaTP/kiota-java/actions

I hope I haven't missed any comment 🙂

baywet:
Thanks for making all those changes. I do think as well you've addressed all my comments, I just resolved them.
Here is a list of comments from other issues we could/should address in this PR as well:

- [ ] #122 (use the maven central task)
- [ ] #119 
- [ ] #35 because of those changes, it gives us an opportunity to simplify the gradlebuild-component, gradlebuild-trigger, and release-component workflows a great deal. They have been problematic in the past and a simplification here would be welcome.

If you don't want to take on the last one because it represents too much work I'd perfectly understand. Let me know what you think

andreaTP:
> Thanks for making all those changes. I do think as well you've addressed all my comments, I just resolved them.

Thanks for the confirmation!
Do you mind if we merge this one and I follow up with separate PRs?
I'm still pretty new to this project and I'll feel more comfortable having incremental changes.

baywet (approved):

PR #125, opened 2023-01-05
Use a single version at the root

### LINKED ISSUE #119: Consider using a single version for all of the components
Hi all, and thanks for the amazing project!

Working with Kiota I'm finding it a bit painful to mix and match the various support libraries versions (abstractions, serialization etc. etc.).
Here I'm proposing to use a single version in this repository and release everything altogether every time.
I understand that this might seem "incorrect" from the semantic point of view, but, since releases are cheap(after the initial setup) I do believe that the benefits are counterbalancing the cons.

Having a single version consistent across the artifacts makes cleaner build files for the users and reduces friction.
e.g. a Maven `pom.xml` might look like this:

```xml
...
  <properties>
    <kiota.version>0.1.0</kiota.version>
  </properties>

  <dependencies>
    <dependency>
      <groupId>com.microsoft.kiota</groupId>
      <artifactId>microsoft-kiota-abstractions</artifactId>
      <version>${kiota.version}</version>
    </dependency>
    <dependency>
      <groupId>com.microsoft.kiota</groupId>
      <artifactId>microsoft-kiota-http-okHttp</artifactId>
      <version>${kiota.version}</version>
    </dependency>
    <dependency>
      <groupId>com.microsoft.kiota</groupId>
      <artifactId>microsoft-kiota-serialization-json</artifactId>
      <version>${kiota.version}</version>
    </dependency>
    <dependency>
      <groupId>com.microsoft.kiota</groupId>
      <artifactId>microsoft-kiota-serialization-text</artifactId>
      <version>${kiota.version}</version>
    </dependency>
    <dependency>
      <groupId>com.microsoft.kiota</groupId>
      <artifactId>microsoft-kiota-serialization-form</artifactId>
      <version>${kiota.version}</version>
    </dependency>
  </dependencies>
  ...
```

andreaTP:
Fix #119 

Proposing to start with `0.1.0` since:
- `0.0` is what has been used so far and we want to mark a change
- the major version is still `0`

I sneaked in some cleanup of the `gradle.properties` file, now the common section is held only in the root properties file.

Question: what to do with the current `CHANGELOG` and `README` files?


baywet:
Thanks for starting this up:
- CHANGELOG: please move the abstractions one to the root, and merge the other ones in it (you can add a -lib name to differentiate), add an entry to the top for a release (set tomorrows date)
- READMEs for components: remove the To-Do, remove the additional repository, update the gradle instruction, a bit like this in essence https://github.com/microsoft/kiota-typescript/tree/main/packages/http/fetch
- README at the root; update the H1 to be the project name (Kiota Java libraries), add an abstract and a using section (if possible for both maven and gradle), leave contributing and trademarks as is. A bit like that https://github.com/microsoft/kiota-abstractions-dotnet Sorry this wasn't done before, this is a miss on our side, if you don't want to do it let me know I'll pick it up.

Let me know if you have additional questions

andreaTP:
> Note: the workflows also need to be refactored

Yup, moved to draft to address this, I'll take care of everything before marking it as ready again 👍 , I need to improve my gradle foo 🙂

.github/workflows/build.yml
```diff
@@ -0,0 +1,40 @@
...
+      with: 
+        java-version: 17
+        distribution: 'adopt'
+        cache: gradle
+    - name: Detect Secrets
+      uses: RobertFischer/detect-secrets-action@v2.0.0
+    - name: Build with Gradle
+      run: ./gradlew --no-daemon build
+    - name: Upload Unit Test Results
+      uses: actions/upload-artifact@v3
```
baywet, line 30:
```suggestion
      if: ${{ always() }}
      uses: actions/upload-artifact@v3
```

we want to upload the test results, especially when they fail

.github/workflows/build.yml
```diff
@@ -0,0 +1,40 @@
...
+on:
+  pull_request:
+    branches: [dev, main]
+    paths-ignore:
+      - '**.gradle/wrapper'
+      - '**.gitignore'
+      - 'LICENSE'
+      - 'THIRD PARTY NOTICES'
+      - '**.md'
+  workflow_dispatch:
```
baywet, line 12:
```suggestion
  workflow_dispatch:
  push:
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,40 @@
+name: Gradle Build Project Trigger
```
baywet, line 1:
```suggestion
name: Gradle Build And Test
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
+name: Build and Publish Trigger
```
baywet, line 1:
```suggestion
name: Gradle Build and Publish
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+        type: string
+      shouldRelease: 
+        required: false
+        type: string
+      environment:
+        required: true
+        type: string        
+      workingDirectory:
+        required: true
+        type: string
```
baywet, line 17:
```suggestion
  push:
    tags: ['v*']
    branches: ['main']
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Download file
+      run: .\scripts\decodeAndWrite.ps1 -encodedValue $env:ENCODED_VALUE -outputPath $env:OUTPUT_PATH
+      shell: pwsh
+      env: 
+        ENCODED_VALUE: ${{ secrets.SECRING_GPG }}
+        OUTPUT_PATH: '.\secring.gpg'
+    - name: Build with Gradle
+      run: ./gradlew --no-daemon build
+    - name: Publish Preview
+      if: ${{ inputs.shouldPreview }} == 'true'
```
baywet, line 51:
```suggestion
      if: github.ref == 'refs/heads/main'
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+      run: ./gradlew --no-daemon $PREVIEW_TASK
+    - name: Publish Release
+      if: ${{ inputs.shouldRelease }} == 'true'
```
baywet, line 54:
```suggestion
      if: contains(github.ref, 'refs/tags/v')
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+      run: ./gradlew --no-daemon $PUBLISH_TASK
```
baywet, line 55:
```suggestion
      run: ./gradlew --no-daemon $PUBLISH_TASK
   - name: Release
     uses: anton-yurchenko/git-release@v5.0
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        DRAFT_RELEASE: "false"
        PRE_RELEASE: "false"
        CHANGELOG_FILE: "CHANGELOG.md"
        ALLOW_EMPTY_CHANGELOG: "true"
```

baywet (changes requested):
Thanks for cleaning this up, here are a couple of suggestions to help with the workflows

.github/workflows/build.yml
andreaTP: omitting the `if` means always AFAIK

.github/workflows/build.yml
andreaTP: yup, adding the `push` with the correct exclusions 👍

.github/workflows/build.yml
baywet: omitting it means only if the previous step was successful, which might not be the case when unit tests fail for instance.

.github/workflows/release.yml
andreaTP: added fixing the yaml indentation

andreaTP:
@baywet you noticed, but this should be ready for review/tweaks until it works as expected 🙂 
You would need to change some settings in GH for required workflows etc.

.github/workflows/build.yml
andreaTP: you are right, thanks for pointing out 👍

.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+    - name: Build with Gradle
+      run: ./gradlew --no-daemon build
+    - name: Publish Preview
+      if: github.ref == 'refs/heads/main'
+      run: ./gradlew --no-daemon $PREVIEW_TASK
+    - name: Publish Release
+      if: contains(github.ref, 'refs/tags/v')
+      run: ./gradlew --no-daemon $PUBLISH_TASK
+    - name: Release
+      uses: anton-yurchenko/git-release@v5.0
```
baywet, line 47:
```suggestion
      if: contains(github.ref, 'refs/tags/v')
      uses: anton-yurchenko/git-release@v5.0
```

baywet (changes requested):
Sorry I missed that part

baywet (approved):
Thanks a lot for cleaning this up! ❤️❤️❤️

.github/workflows/release.yml
```diff
@@ -0,0 +1,54 @@
...
+    tags: ['v*']
+    branches: ['main']
+
+env:
+  PREVIEW_TASK: publishSnapshotPublicationToSonatypeSnapshotRepository
+  PUBLISH_TASK: publishMavenCentralReleasePublicationToSonatypeRepository
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
baywet, line 15:
```suggestion
    runs-on: ubuntu-latest
    env: maven_central
```

baywet (changes requested):

baywet (approved):

baywet (approved):

PR #126, opened 2023-01-05
feature/user agent

baywet:
related https://github.com/microsoft/kiota/issues/2056

### DIFF
--- a/components/http/okHttp/src/main/java/com/microsoft/kiota/http/middleware/options/UserAgentHandlerOption.java
+++ b/components/http/okHttp/src/main/java/com/microsoft/kiota/http/middleware/options/UserAgentHandlerOption.java
@@ -12,7 +12,7 @@ public class UserAgentHandlerOption implements RequestOption {
 	@Nonnull
 	private String productName = "kiota-java";
 	@Nonnull
-	private String productVersion = "0.0.10";
+	private String productVersion = "0.1.1";
 	/**
 	 * Gets the product name to be used in the user agent header
 	 * @return the product name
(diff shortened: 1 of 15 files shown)


components/http/okHttp/src/main/java/com/microsoft/kiota/http/middleware/options/UserAgentHandlerOption.java
```diff
@@ -0,0 +1,69 @@
...
+
+import com.microsoft.kiota.RequestOption;
+
+/** The options for the user agent handler */
+public class UserAgentHandlerOption implements RequestOption {
+	private boolean enabled = true;
+	@Nonnull
+	private String productName = "kiota-java";
+	@Nonnull
+	private String productVersion = "0.0.10";
```
andrueastman, line 15: Is it possible to read this at runtime? Otherwise we may have to keep updating it as well

baywet: Not as far as my research yielded. The version number is merely metadata on the repository (package feed) but not part of the jar itself like it is in dotnet.

baywet: Actually this might be worth trying https://stackoverflow.com/a/21725706

baywet: after trying it, it will only work when packaged as a jar and added to the project from the feed (with its manifest). That's already a problem in itself for the unit test. But there's also a packaging technique called fat jars in Java where people mush all the dependencies and their code in a single jar, loosing all the manifests in the process.
This means the package version would be outside of our control.

andrueastman: Thanks for clearing this one up. Resolving for now.

andrueastman (approved):

PR #138, opened 2023-01-09
feature/native request

baywet:
related https://github.com/microsoft/kiota/issues/2084


baywet:
@andreaTP as you can see here the CI fails when there's a denpendency between a change in implementations and a change in abstractions. I wonder whether we could fix that aspect while automating the reference update bump. Thoughts?

andrueastman (approved):

PR #139, opened 2023-01-09
Declare interdependencies as project deps

andreaTP:


baywet (approved):
Thanks a lot!
I trust that during publishing gradle will use the package names and versions?

andreaTP:
@baywet Sorry for the noise, I just found that this PR likely broke the publishing mechanism 🙁 
You can reproduce locally with `./gradlew publishToMavenLocal`

andreaTP:
@baywet looks like I got into a rabbit hole here, please revert and let's try to fix everything in another PR.

baywet:
I just fixed it with #140 I believe. Have a look and let me know. We didn't need any of the publications except for one.

PR #140, opened 2023-01-09
- removes unecessary publication definitions

baywet:
related #139, CC @AndreaTP

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,7 @@ on:
     branches: ['main']
 
 env:
-  PREVIEW_TASK: publishSnapshotPublicationToSonatypeSnapshotRepository
+  PREVIEW_TASK: publishMavenPublicationToSonatypeSnapshotRepository
   PUBLISH_TASK: publishmavenPublicationToSonatypeRepository
 
 jobs:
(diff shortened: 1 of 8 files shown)


.github/workflows/release.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 env:
   PREVIEW_TASK: publishSnapshotPublicationToSonatypeSnapshotRepository
```
baywet, line 10:
```suggestion
  PREVIEW_TASK: publishMavenPublicationToSonatypeSnapshotRepository
```

andreaTP:
This sounds like a great improvement!
It's not clear to me how those configurations are used ...

baywet:
> It's not clear to me how those configurations are used
if the comment was about where they might be used, here is a bit more context:
- Snapshot is a remnant from msgraph-java-sdk-core, it's effectively only here to add the snapshot suffix, which can be passed by the CLI.
- Maven (the one adding a few properties) is a remnant from when we used to publish to a GitHub packages repository.

neither are needed anymore AFAIK, hence their removal.

baywet:
> use the final version number only for Maven Central releases

> Since you are at it I try, it would be great if the version is computed as follows:
> 
> * always use the suffix `-SNAPSHOT` for development/local releases
> * use the final version number only for Maven Central releases
> 
> This is going to play much better with the rest of the tools/build systems than the current setup.

@andreaTP done, thanks for the suggestion. Let me know if you think we should change anything else

andreaTP (approved):
LGTM, great cleanup, thanks!

PR #141, opened 2023-01-09
Fix publish to Maven Local

andreaTP:
This seems to fix the issue with multiple publications @baywet 

But I really encourage you to double check if we are not breaking something as it looks to be a rough edge of Gradle itself.


baywet (changes requested):
sorry for the confusion here, blocking for now until we review #140 which has the benefit of simplifying further the configuration of the project

andreaTP:
Yup, no problem, I let you go 👍

PR #268, opened 2023-04-05
getValue should be @nonNull

ramsessanchez:
Quick annotation fix, no bump necessary.


components/abstractions/src/main/java/com/microsoft/kiota/NativeResponseHandler.java
```diff
@@ -39,7 +39,7 @@ public Map<String, ParsableFactory<? extends Parsable>> getErrorMappings() {
      * Get the value for the response.
      * @return the value of the response.
      */
-    @Nullable
+    @Nonnull
```
baywet, line 42: this is technically not set by the the constructor and can be null.

baywet (changes requested):

PR #266, opened 2023-03-30
Native response handler

ramsessanchez:
Adds a native response handler for use in the LargeFileUpload task requests used in Java-Core for Kiota. 
Adds setResponseHandler function to the requestInformation class.

PR #298, opened 2023-04-13
- upgrades to gadle eight to unblock android plugin PRs

baywet:
follow up to #266 and #267
closes #297, #296, #295, #294, #293, #292


andrueastman (approved):
👍🏼

PR #299, opened 2023-04-17
Adds reponseheaders to API excpetion

andrueastman:
Related to https://github.com/microsoft/kiota/issues/2524

### DIFF
--- a/components/abstractions/src/main/java/com/microsoft/kiota/ApiException.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/ApiException.java
@@ -1,5 +1,7 @@
 package com.microsoft.kiota;
 
+import java.util.List;
+import java.util.Map;
 import javax.annotation.Nonnull;
 
 /** Parent type for exceptions thrown by the client when receiving failed responses to its requests. */
@@ -23,4 +25,7 @@ public class ApiException extends Exception {
 
     /** The HTTP status code  for the response*/
     public int responseStatusCode;
+
+    /** The HTTP response headers for the error response*/
+    public Map<String, List<String>> responseHeaders;
 }
--- a/components/abstractions/src/main/java/com/microsoft/kiota/ApiException.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/ApiException.java
@@ -1,5 +1,6 @@
 package com.microsoft.kiota;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nonnull;
@@ -27,5 +28,6 @@ public class ApiException extends Exception {
     public int responseStatusCode;
 
     /** The HTTP response headers for the error response*/
-    public Map<String, List<String>> responseHeaders;
+    @Nonnull
+    public Map<String, List<String>> responseHeaders = new HashMap<String, List<String>>();
 }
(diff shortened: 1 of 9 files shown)


components/abstractions/src/main/java/com/microsoft/kiota/ApiException.java
```diff
@@ -23,4 +26,8 @@ public ApiException(@Nonnull final Throwable cause) {
 
     /** The HTTP status code  for the response*/
     public int responseStatusCode;
+
+    /** The HTTP response headers for the error response*/
+    @Nonnull
+    public Map<String, List<String>> responseHeaders = new HashMap<String, List<String>>();
```
baywet, line 32: The keys should be case insensitive

baywet (dismissed):

components/abstractions/src/main/java/com/microsoft/kiota/ApiException.java
```diff
@@ -23,4 +26,8 @@ public ApiException(@Nonnull final Throwable cause) {
 
     /** The HTTP status code  for the response*/
     public int responseStatusCode;
+
+    /** The HTTP response headers for the error response*/
+    @Nonnull
+    public ResponseHeaders responseHeaders = new ResponseHeaders();
```
baywet, line 32:
```suggestion
    public final ResponseHeaders responseHeaders = new ResponseHeaders();
```

baywet (changes requested):

baywet (approved):

components/abstractions/src/main/java/com/microsoft/kiota/Headers.java
```diff
@@ -0,0 +1,53 @@
+package com.microsoft.kiota;
+
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
+
+import javax.annotation.Nonnull;
+
+class Headers extends CaseInsensitiveMap {
```
baywet, line 9: we should either:
- have a copy method
- in API exception, have getters and setters and check that nobody is assigning null
Otherwise there's no way we can enforce the nonnull attribute without the final modifier (my last commit)

components/abstractions/src/main/java/com/microsoft/kiota/ApiException.java
```diff
@@ -29,5 +28,17 @@ public ApiException(@Nonnull final Throwable cause) {
 
     /** The HTTP response headers for the error response*/
     @Nonnull
-    public final ResponseHeaders responseHeaders = new ResponseHeaders();
+    private ResponseHeaders responseHeaders = new ResponseHeaders();
+
+    /** Gets the HTTP response headers for the error response */
+    public ResponseHeaders getResponseHeaders() {
+        return new ResponseHeaders(responseHeaders);
```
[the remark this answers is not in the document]
andrueastman: Yes that would be case. 
Spotbugs raises this as a flag/bug if we assign directly.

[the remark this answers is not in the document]
baywet: Ah right they don't want people to be able to mutate the underlying object with an external reference.
I don't expect anyone will try to set those besides our own code, let's roll with that and see what kind of feedback we get.

baywet (approved):

PR #517, opened 2023-07-26
feature/multipart

baywet:
related https://github.com/microsoft/kiota/issues/220

### DIFF
--- a/components/abstractions/src/main/java/com/microsoft/kiota/MultipartBody.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/MultipartBody.java
@@ -68,13 +68,16 @@ public class MultipartBody implements Parsable {
 	 * @return the content type of the part with the specified name.
 	 */
 	@Nullable
-	public Map.Entry<String, Object> getPartValue(@Nonnull final String partName)
+	public Object getPartValue(@Nonnull final String partName)
 	{
 		Objects.requireNonNull(partName);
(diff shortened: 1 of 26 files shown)


components/abstractions/src/main/java/com/microsoft/kiota/MultipartBody.java
```diff
@@ -0,0 +1,171 @@
...
+	 * @param value the value of the part to add or replace.
+	 */
+	public <T> void addOrReplacePart(@Nonnull final String name, @Nonnull final String contentType, @Nonnull final T value) {
+		Objects.requireNonNull(name);
+		Objects.requireNonNull(contentType);
+		Objects.requireNonNull(value);
+		if (contentType.isBlank() || contentType.isEmpty())
+			throw new IllegalArgumentException("contentType cannot be blank or empty");
+		if (name.isBlank() || name.isEmpty())
+			throw new IllegalArgumentException("name cannot be blank or empty");
```
ramsessanchez, line 53: Might be a good idea to leverage googles String library and use
Strings.IsNullOrEmpty(name) and consolidate the illegalArgumentExceptions with the requireNonNull exceptions.

baywet: Excellent point! updated.

andrueastman (approved):

PR #553, opened 2023-08-08
Jakarta Annotations

ramsessanchez:
Replaces Javax annotations for Kiota-Java Components in favor of Jakarta annotations.

### DIFF
--- a/components/abstractions/gradle/dependencies.gradle
+++ b/components/abstractions/gradle/dependencies.gradle
@@ -10,7 +10,7 @@ dependencies {
     // This dependency is used internally, and not exposed to consumers on their own compile classpath.
     implementation 'com.google.guava:guava:32.1.2-jre'
     implementation 'org.javatuples:javatuples:1.2'
-    implementation 'javax.annotation:javax.annotation-api:1.3.2'
+    implementation 'jakarta.annotation:jakarta.annotation-api:2.1.1'
     implementation 'io.opentelemetry:opentelemetry-api:1.28.0'
     implementation 'io.opentelemetry:opentelemetry-context:1.28.0'
 }
--- a/components/abstractions/gradle/dependencies.gradle
+++ b/components/abstractions/gradle/dependencies.gradle
@@ -7,12 +7,10 @@ dependencies {
     // Use JUnit Jupiter Engine for testing.
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
 
-    //This dependency is used internally and also by Kiota generated code which leverages this library.
-    api 'jakarta.annotation:jakarta.annotation-api:2.1.1'
-
     // This dependency is used internally, and not exposed to consumers on their own compile classpath.
     implementation 'com.google.guava:guava:32.1.2-jre'
     implementation 'org.javatuples:javatuples:1.2'
     implementation 'io.opentelemetry:opentelemetry-api:1.28.0'
     implementation 'io.opentelemetry:opentelemetry-context:1.28.0'
+    implementation 'jakarta.annotation:jakarta.annotation-api:2.1.1'
 }
--- a/components/abstractions/gradle/dependencies.gradle
+++ b/components/abstractions/gradle/dependencies.gradle
@@ -7,10 +7,12 @@ dependencies {
     // Use JUnit Jupiter Engine for testing.
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
 
+    //This dependency is used internally and also by Kiota generated code which leverages this library.
+    api 'jakarta.annotation:jakarta.annotation-api:2.1.1'
+
     // This dependency is used internally, and not exposed to consumers on their own compile classpath.
     implementation 'com.google.guava:guava:32.1.2-jre'
     implementation 'org.javatuples:javatuples:1.2'
-    implementation 'jakarta.annotation:jakarta.annotation-api:2.1.1'
     implementation 'io.opentelemetry:opentelemetry-api:1.28.0'
     implementation 'io.opentelemetry:opentelemetry-context:1.28.0'
 }
--- a/components/serialization/form/src/test/java/com/microsoft/kiota/serialization/mocks/TestEntity.java
+++ b/components/serialization/form/src/test/java/com/microsoft/kiota/serialization/mocks/TestEntity.java
@@ -137,8 +137,8 @@ public class TestEntity implements Parsable, AdditionalDataHolder {
 	public Map<String, Object> getAdditionalData() {
 		return _additionalData;
 	}
-	@javax.annotation.Nonnull
-    public static TestEntity createFromDiscriminatorValue(@javax.annotation.Nonnull final ParseNode parseNode) {
+	@jakarta.annotation.Nonnull
+    public static TestEntity createFromDiscriminatorValue(@jakarta.annotation.Nonnull final ParseNode parseNode) {
 		return new TestEntity();
 	}
 }
--- a/components/serialization/form/src/test/java/com/microsoft/kiota/serialization/mocks/TestEntity.java
+++ b/components/serialization/form/src/test/java/com/microsoft/kiota/serialization/mocks/TestEntity.java
@@ -54,7 +54,7 @@ public class TestEntity implements Parsable, AdditionalDataHolder {
 		return _workDuration;
 	}
 
-	public void setWorkDuration(final PeriodAndDuration value) {
+	public void setWorkDuration(PeriodAndDuration value) {
 		this._workDuration = PeriodAndDuration.of(value.getPeriod(), value.getDuration());
 	}
 
--- a/components/serialization/form/src/test/java/com/microsoft/kiota/serialization/mocks/TestEntity.java
+++ b/components/serialization/form/src/test/java/com/microsoft/kiota/serialization/mocks/TestEntity.java
@@ -54,8 +54,8 @@ public class TestEntity implements Parsable, AdditionalDataHolder {
 		return _workDuration;
 	}
 
-	public void setWorkDuration(PeriodAndDuration value) {
-		this._workDuration = value;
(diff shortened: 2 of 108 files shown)


andrueastman (dismissed):

baywet (changes requested):
Thanks for starting this.
We still have an immutable annotation from javax https://github.com/microsoft/kiota-java/blob/dd9d15bf37c5a32d005f53b358575b043bd0682f/components/abstractions/src/main/java/com/microsoft/kiota/PeriodAndDuration.java#L4

Which [doesn't seem to have an equivalent with jakarta for some reason](https://github.com/jakartaee/common-annotations-api/issues/118) we should try to find an equivalent. Maybe @andreaTP can help with this?

andreaTP:
@baywet is this annotation effectively used by one or more compile/runtime components?

I think [this SO](https://stackoverflow.com/a/27110465) answer is pretty relevant here.

components/abstractions/gradle/dependencies.gradle
```diff
@@ -10,7 +10,7 @@ dependencies {
     // This dependency is used internally, and not exposed to consumers on their own compile classpath.
     implementation 'com.google.guava:guava:32.1.2-jre'
     implementation 'org.javatuples:javatuples:1.2'
-    implementation 'javax.annotation:javax.annotation-api:1.3.2'
+    implementation 'jakarta.annotation:jakarta.annotation-api:2.1.1'
```
baywet, line 13: we might want to switch this one to API instead so consumers get the dependency automatically for the generated code. What do you think @andreaTP ?

ramsessanchez: This would also mean that we don't have to specify the dependency for the other components since the abstractions would make it available for them.

andreaTP:
In this case I would say that it's safe to remove the annotation.

baywet:
@andreaTP Ramses was able to identify a replacement com.google.errorprone.annotations.Immutable, we already have that dependency here so it's a easy swap.

components/abstractions/gradle/dependencies.gradle
andreaTP: @baywet I'm not sure I'm following here, at the moment the dependency is on API afaiu, or am I missing something?

baywet: correct, I know we've already had dependency management issues in the past. This is why I'm careful about making it an API dependency and would like @andreaTP point of view.
The alternative being to include the dependency in the kiota info command. It be a first since we don't require people to include anything else besides abstractions.

baywet: @anreaTp I guess we were writing at the same time. Because we depend on that annotation for the generated code, people would need to add it to there project dependency for things to compile if we leave that as an implementation dependency.

andreaTP:
Imposing errorprone to users is a big toll imho, I would not be in favor of expanding it's usage, especially if explicit.

components/abstractions/gradle/dependencies.gradle
andreaTP: @baywet I'm about to log off for today, I think that what you are saying is not accurate.
At the moment we do ask the user to provide the findbugs dependency(at least for compile time).

I'm completely in favor of leaving this to the user, and, possibly, adding it to `kiota info` as you proposed.
Compile time dependency should not spill to the user runtime and the user should be in control of it.

baywet:
alright I guess we can just remove it then. CC @ramsessanchez (immutable)

components/serialization/form/src/test/java/com/microsoft/kiota/serialization/mocks/TestEntity.java
```diff
@@ -54,8 +54,8 @@ public PeriodAndDuration getWorkDuration() {
 		return _workDuration;
 	}
 
-	public void setWorkDuration(PeriodAndDuration value) {
-		this._workDuration = value;
+	public void setWorkDuration(final PeriodAndDuration value) {
```
andrueastman, line 57: Just to confirm here, if this change is a side effect of removing the immutable annotation, will we need a corresponding change in the generation logic to avoid a warning/error?

andreaTP: As far as I understand it, this change is not needed.

components/serialization/json/spotBugsExcludeFilter.xml
```diff
@@ -26,4 +26,8 @@ xsi:schemaLocation="https://github.com/spotbugs/filter/3.0.0 https://raw.githubu
 			<Class name="com.microsoft.kiota.serialization.UnionWrapperParseTests$4$1" />
 		</Or>
 	</Match>
+	<Match>
+		<Bug pattern="NP_BOOLEAN_RETURN_NULL" />
```
andreaTP, line 30: should we fix this issue instead of excluding it?

gradle.properties
```diff
@@ -25,7 +25,7 @@ org.gradle.caching=true
 
 mavenGroupId         = com.microsoft.kiota
 mavenMajorVersion = 0
```
andreaTP, line 27: this change is breaking, we should be bumping the major version IMHO

andreaTP (approved):
A couple of minor comments but, overall, LGTM, thanks @ramsessanchez for the effort!

gradle.properties
baywet: no, we're still not stable for the Java language, so breaking changes are still permitted as per our support policy

components/abstractions/gradle/dependencies.gradle
baywet: The different being [nullable/nonnull are runtime annotations](https://github.com/jakartaee/common-annotations-api/blob/0e6318ac2716c463b33cceff191c9639c73e80ca/api/src/main/java/jakarta/annotation/Nullable.java#L41) and findbugs [is a class annotation](https://github.com/spotbugs/spotbugs/blob/2fdbb01514683c32c07aeb841879269dddccbafc/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/SuppressFBWarnings.java#L33) so it gets erased earlier in the lifecycle.

We're getting rid of the suppress fb warnings annotations to trim down our dependencies (or the ones we're pushing on the consumer). So this won't be a problem for long.

Let's play the worst case scenarios for both options:
- implementation, people don't add it to their dependencies, I'm guessing they'd get an error at compile time saying they are missing a dependency?
- api, people have a different version brought by another dependency, now they are stuck with a conflict that's hard to resolve

Am I understanding this correctly?

andreaTP: I completely switch api/implementation with the respective worst case 🙂 .
Let's talk about the concrete example:

`jakarta.annotation:jakarta.annotation-api:2.1.1`

This is shipping the API but not the implementation, people will be asked to provide it on the classpath.
I think that keeping this is perfectly fine.

baywet: alright @ramsessanchez can you add that dependency to the appsettings in your kiota PR please?
Thanks andrea for the help here

components/serialization/form/src/test/java/com/microsoft/kiota/serialization/mocks/TestEntity.java
ramsessanchez: The 'final' is not needed, however the way that the workDuration values is set
``` Java 
this._workDuration = PeriodAndDuration.of(value.getPeriod(), value.getDuration());
```
Will likely need to change in generation. 
@baywet , @andrueastman , do we feel this is a fair pattern to follow in generation?

components/serialization/json/spotBugsExcludeFilter.xml
ramsessanchez: I believe this is an intended implementation, though I think a comment should be added to explain the reasoning. 
``` Java
    @Nullable
    public Boolean getBooleanValue() {
        return currentNode.isJsonPrimitive() ? currentNode.getAsBoolean() : null;
    }
```
Always returning a boolean in this above method would suggest that the parseNode has a valid boolean value, which, if the node is not a jsonPrimitive would not be true. By retuning 'null' we avoid throwing an exception and rather return 'null' to suggest that it does not have a valid boolean value. The Nullable annotation is there to document that this method may return a null value, however SpotBugs continues to flag this method. My opinion is that a comment should be added to the getBooleanValue method in the ParseNode.java interface within the abstraction library to communicate the meaning of a null value.

components/serialization/form/src/test/java/com/microsoft/kiota/serialization/mocks/TestEntity.java
andrueastman: @ramsessanchez This looks good to me. I understand the need to do this as the removal of the `Immutable` annotation leads to the analyzers causing build error.

components/abstractions/gradle/dependencies.gradle
baywet: yep, leave implementation as is, add an entry in the appsettings.json file in kiota. Thanks

components/abstractions/gradle/dependencies.gradle
```diff
@@ -7,10 +7,12 @@ dependencies {
     // Use JUnit Jupiter Engine for testing.
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
 
+    //This dependency is used internally and also by Kiota generated code which leverages this library.
+    api 'jakarta.annotation:jakarta.annotation-api:2.1.1'
```
baywet, line 11:
```suggestion
    implementation 'jakarta.annotation:jakarta.annotation-api:2.1.1'
```
I thought we had agreed to leave as implementation and add the dependency in the app settings? Cc @andreaTP

ramsessanchez: Ah yes, I reverted the changes and forgot to push.

baywet (approved):

PR #575, opened 2023-08-14
- adds headers inspection middleware

baywet:
partial https://github.com/microsoft/kiota/issues/2963


ramsessanchez (approved):
note to add a value to the telemetry for this handler to track its usage.

PR #602, opened 2023-09-02
#585 Prevent decoding of parameter values in the ParametersNameDecodingHandler Interceptor & write corresponding test case

### LINKED ISSUE #585: parameters names decoding handler also decodes values
originally reported in https://github.com/microsoft/kiota-http-python/issues/179
We add those cases as unit tests and correct the implementation so it doesn't impact query parameters values

Avzel:
This pull request fixes #585, ensuring that the ParametersNameDecodingHandler only decodes query parameter names (and not values). I wrote a test case using the problem URLs that were linked in the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+## [0.7.3] - 2023-09-04
+
+### Fixed
+
+- Fixed bug that caused the ParametersNameDecodingHandler to decode query parameter values in addition to names
+
 ## [0.7.2] - 2023-09-01
 
 ### Changed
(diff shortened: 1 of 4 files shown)


Avzel:

andrueastman (commented):
Thanks for the contribution here @Avzel 

Any chance change you can bump up the version number in the `gradle.properties` file and add a changelog entry in `changelog.md` before we merge?

Avzel:
> Any chance change you can bump up the version number in the `gradle.properties` file and add a changelog entry in `changelog.md` before we merge?

Sure not a problem, I just committed those updates

andrueastman (approved):
Thanks again for the contribution @Avzel

baywet:
Thanks for the contribution @Avzel ! publishing now

PR #674, opened 2023-10-03
Add tryAdd method to headers

andreaTP:
Preparation for https://github.com/microsoft/kiota/pull/3401

### DIFF
--- a/components/abstractions/src/main/java/com/microsoft/kiota/Headers.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/Headers.java
@@ -31,18 +31,35 @@ public abstract class Headers extends CaseInsensitiveMap {
      * @param value the value of the header to add.
      */
     public void add(@Nonnull final String key, @Nonnull final String value) {
+        addImpl(key, value, true);
+    }
+
+    /**
+     * Adds a header to the current request if it was not already set
+     *
+     * @param key   the key of the header to add.
+     * @param value the value of the header to add.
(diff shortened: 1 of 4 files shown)


components/abstractions/src/main/java/com/microsoft/kiota/Headers.java
```diff
@@ -31,19 +31,36 @@ protected Headers(@Nonnull Headers headers) {
...
+        addImpl(key, value, true);
+    }
+
+    /**
+     * Adds a header to the current request if it was not already set
+     *
+     * @param key   the key of the header to add.
+     * @param value the value of the header to add.
+     */
+    public void tryAdd(@Nonnull final String key, @Nonnull final String value) {
```
baywet, line 43: maybe this should return a boolean to indicate whether or not the add was successful? (at least it's done this way in dotnet)

baywet (changes requested):
Can you add a changelog entry please (bump minor, set the date to tomorrow)
And update gradle.properties with the new version.

andreaTP:
Sure! Done! 👍

baywet (approved):
Changelog change was done in #675 which is fine.

andreaTP:
> Changelog change was done in https://github.com/microsoft/kiota-java/pull/675 which is fine.

I reverted it there as there is still something to fix for the linter, applied it here though

baywet (approved):

PR #675, opened 2023-10-03
ApiException mutable field better encapsulation

### LINKED ISSUE #654: [Questions] On `ApiException` design
Hi there 👋 !

I'm digging down a bit in the codebase and I have a couple of questions regarding the design of `ApiException` and how you envision its interaction with Authentication mechanisms:

1. [`requestStatusCode`](https://github.com/microsoft/kiota-java/blob/f41fcfefa7eb119f1e680b0e1bca3cd36c11660b/components/abstractions/src/main/java/com/microsoft/kiota/ApiException.java#L27) is a public mutable field, which seems odd, once the Exception is thrown we want the `statusCode` to be a stable number that cannot be easily changed in user code.
2. `ApiException` is a ["checked exception"](https://github.com/microsoft/kiota-java/blob/f41fcfefa7eb119f1e680b0e1bca3cd36c11660b/components/abstractions/src/main/java/com/microsoft/kiota/ApiException.java#L8) which means that it should be explicitly written in all the method signatures using it. There are quite some discussions around checked exceptions in Java, but, in this context, how do I throw an `ApiException(401)` from an `AuthenticationProvider` so that it's indistinguishable from a business logic 401 handled at the API level?

andreaTP:
fixes #654 , using a builder pattern we can better encapsulate the `responseStatusCode` mutable field.

I think it's a code smell to pass to the user-land an object with a mutable field, since we have all of the information to build the final one it should not be exposed to the user.

### DIFF
--- a/components/abstractions/src/main/java/com/microsoft/kiota/ApiException.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/ApiException.java
@@ -24,7 +24,25 @@ public class ApiException extends Exception {
     }
 
     /** The HTTP status code  for the response*/
-    public int responseStatusCode;
+    private int responseStatusCode;
+
+    /**
+     * Gets the HTTP response status code
+     * @return The response status code from the failed response.
+     */
+    @Nonnull
+    public int getResponseStatusCode() {
+        return responseStatusCode;
+    }
+
+    /**
+     * Sets the HTTP response status code
+     * @param responseStatusCode The response status code to set.
+     */
+    protected void setResponseStatusCode(int responseStatusCode) {
+        Objects.requireNonNull(responseStatusCode);
+        this.responseStatusCode = responseStatusCode;
+    }
 
     /** The HTTP response headers for the error response*/
     @Nonnull
@@ -43,7 +61,7 @@ public class ApiException extends Exception {
      * Sets the HTTP response headers for the error response
      * @param responseHeaders The response headers collections to set.
      */
-    public void setResponseHeaders(@Nonnull ResponseHeaders responseHeaders) {
+    protected void setResponseHeaders(@Nonnull ResponseHeaders responseHeaders) {
         Objects.requireNonNull(responseHeaders);
         this.responseHeaders = new ResponseHeaders(responseHeaders);
     }
--- a/components/abstractions/src/main/java/com/microsoft/kiota/ApiException.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/ApiException.java
@@ -40,7 +40,6 @@ public class ApiException extends Exception {
      * @param responseStatusCode The response status code to set.
      */
     protected void setResponseStatusCode(int responseStatusCode) {
-        Objects.requireNonNull(responseStatusCode);
         this.responseStatusCode = responseStatusCode;
     }
 
--- a/components/abstractions/src/main/java/com/microsoft/kiota/ApiExceptionBuilder.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/ApiExceptionBuilder.java
@@ -29,6 +29,7 @@ public class ApiExceptionBuilder {
      * @param message The message to be attached to this ApiException.
      * @return The builder object.
      */
+    @Nonnull
     public ApiExceptionBuilder withMessage(@Nonnull String message) {
         Objects.requireNonNull(message);
         if (value == null) {
@@ -44,6 +45,7 @@ public class ApiExceptionBuilder {
      * @param exception The Throwable to be used as Cause for this ApiException.
      * @return The builder object.
      */
+    @Nonnull
     public ApiExceptionBuilder withThrowable(@Nonnull Throwable exception) {
         Objects.requireNonNull(exception);
         if (value == null) {
@@ -59,6 +61,7 @@ public class ApiExceptionBuilder {
      * @param responseStatusCode an int representing the response status code.
      * @return The builder object.
      */
+    @Nonnull
     public ApiExceptionBuilder withResponseStatusCode(int responseStatusCode) {
         if (value == null) {
             value = new ApiException();
@@ -72,6 +75,7 @@ public class ApiExceptionBuilder {
      * @param responseHeaders the response headers to be added to this ApiException.
      * @return The builder object.
      */
+    @Nonnull
     public ApiExceptionBuilder withResponseHeaders(@Nonnull ResponseHeaders responseHeaders) {
         if (value == null) {
             value = new ApiException();
@@ -84,6 +88,7 @@ public class ApiExceptionBuilder {
      * Build and return an instance of ApiException
      * @return The built ApiException.
      */
+    @Nonnull
     public ApiException build() {
         ApiException result = value;
         value = null;
--- a/components/abstractions/src/main/java/com/microsoft/kiota/ApiExceptionBuilder.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/ApiExceptionBuilder.java
@@ -1,0 +1,51 @@
+package com.microsoft.kiota;
+
+import jakarta.annotation.Nonnull;
+
+import java.util.Objects;
+
+/** Builder class for ApiException. */
+public class ApiExceptionBuilder {
+
+    private ApiException value = null;
+    public ApiExceptionBuilder() {
+    }
+    public ApiExceptionBuilder(ApiException base) {
+        value = base;
+    }
+    public ApiExceptionBuilder withMessage(@Nonnull String message) {
+        Objects.requireNonNull(message);
+        if (value == null) {
+            value = new ApiException(message);
+        } else {
+            value = new ApiException(message, value.getCause());
+        }
+        return this;
+    }
+    public ApiExceptionBuilder withThrowable(@Nonnull Throwable exception) {
+        Objects.requireNonNull(exception);
+        if (value == null) {
+            value = new ApiException(exception);
+        } else {
+            value = new ApiException(value.getMessage(), value.getCause());
+        }
+        return this;
+    }
+    public ApiExceptionBuilder withResponseStatusCode(int responseStatusCode) {
+        if (value == null) {
+            value = new ApiException();
+        }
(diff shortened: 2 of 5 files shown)


components/abstractions/src/main/java/com/microsoft/kiota/ApiException.java
```diff
@@ -24,7 +24,25 @@ public ApiException(@Nonnull final Throwable cause) {
...
+    public int getResponseStatusCode() {
+        return responseStatusCode;
+    }
+
+    /**
+     * Sets the HTTP response status code
+     * @param responseStatusCode The response status code to set.
+     */
+    protected void setResponseStatusCode(int responseStatusCode) {
+        Objects.requireNonNull(responseStatusCode);
```
baywet, line 43:
```suggestion
```
int's can't be null? or does this method also check for 0 ?

components/abstractions/src/main/java/com/microsoft/kiota/ApiExceptionBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.Objects;
+
+/** Builder class for ApiException. */
+public class ApiExceptionBuilder {
+
+    private ApiException value = null;
+    public ApiExceptionBuilder() {
+    }
+    public ApiExceptionBuilder(ApiException base) {
```
baywet, line 13: all methods in this class are missing doc comments

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
```diff
@@ -611,24 +614,29 @@ private Response throwIfFailedResponse(@Nonnull final Response response, @Nonnul
...
                 }
                 spanForAttributes.setAttribute(errorBodyFoundAttributeName, true);
                 final Span deserializationSpan = GlobalOpenTelemetry.getTracer(obsOptions.getTracerInstrumentationName()).spanBuilder("getObjectValue").setParent(Context.current().with(span)).startSpan();
                 try(final Scope deserializationScope = deserializationSpan.makeCurrent()) {
                     final Parsable error = rootNode.getObjectValue(errorClass);
-                    ApiException result;
+                    ApiExceptionBuilder resultBuilder;
                     if (error instanceof ApiException) {
-                        result = (ApiException)error;
+                        resultBuilder = new ApiExceptionBuilder((ApiException)error);
```
baywet, line 631: I'm not sure about that last change. We'd effectively be wrapping all exceptions in a parent Api exception. Since all exceptions inherit from API exception, this seems superfluous, and will create confusion: which object contains the headers and status code from a consumer perspective?

components/abstractions/src/main/java/com/microsoft/kiota/ApiException.java
andreaTP: no they cannot be null you are right

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
andreaTP: > We'd effectively be wrapping all exceptions in a parent Api exception.

why? here we only pass through a Builder pattern to keep the final object immutable for the user, that's it.
This change is mostly cosmetics and doesn't alter the original behavior (AFAIU).

> which object contains the headers and status code from a consumer perspective?

The final `ApiException` returned to the user.

Have I missed any invariant here?

components/abstractions/src/main/java/com/microsoft/kiota/ApiExceptionBuilder.java
andreaTP: done

components/abstractions/src/main/java/com/microsoft/kiota/ApiExceptionBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Constructs an empty ApiExceptionBuilder
+     */
+    public ApiExceptionBuilder() {
+    }
+
+    /**
+     * Constructs an ApiExceptionBuilder starting from a base ApiException
+     */
+    public ApiExceptionBuilder(ApiException base) {
```
baywet, line 21: missing doc comment for the parameter

components/abstractions/src/main/java/com/microsoft/kiota/ApiExceptionBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+        value = base;
+    }
+
+    /**
+     * Assign the message to the builder
+     * @return The builder object.
+     */
+    public ApiExceptionBuilder withMessage(@Nonnull String message) {
```
baywet, line 29: missing doc comment for the parameter (for all methods)

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
baywet: right, I missed a subtility in the builder implementation. Disregard.

baywet (changes requested):

components/abstractions/src/main/java/com/microsoft/kiota/ApiExceptionBuilder.java
andreaTP: done, hopefully 🤞

baywet (dismissed):

baywet (approved):

andreaTP:
Fixed, if you want the Changelog entry, just merge first the other PR and I'll follow up.

andreaTP:
should be ready now 👍

baywet (dismissed):

components/abstractions/src/main/java/com/microsoft/kiota/ApiExceptionBuilder.java
```diff
@@ -0,0 +1,91 @@
...
+     * Constructs an empty ApiExceptionBuilder
+     */
+    public ApiExceptionBuilder() {
+    }
+
+    /**
+     * Constructs an ApiExceptionBuilder starting from a base ApiException
+     * @param base The original ApiException to be used as a base.
+     */
+    public ApiExceptionBuilder(ApiException base) {
```
baywet, line 22:
```suggestion
    public ApiExceptionBuilder(@Nonnull final ApiException base) {
        Objects.requireNonNull(base);
```

baywet (approved):

components/abstractions/src/main/java/com/microsoft/kiota/ApiExceptionBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+        Objects.requireNonNull(base);
+        value = new ApiException(base.getMessage(), base.getCause());
+    }
+
+    /**
+     * Assign the message to the builder
+     * @param message The message to be attached to this ApiException.
+     * @return The builder object.
+     */
+    public ApiExceptionBuilder withMessage(@Nonnull String message) {
```
baywet, line 32:
```suggestion
    @Nonnull
    public ApiExceptionBuilder withMessage(@Nonnull String message) {
```

components/abstractions/src/main/java/com/microsoft/kiota/ApiExceptionBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+        return this;
+    }
+
+    /**
+     * Assign the Throwable cause of the Exception to the builder
+     * @param exception The Throwable to be used as Cause for this ApiException.
+     * @return The builder object.
+     */
+    public ApiExceptionBuilder withThrowable(@Nonnull Throwable exception) {
```
baywet, line 47:
```suggestion
    @Nonnull
    public ApiExceptionBuilder withThrowable(@Nonnull Throwable exception) {
```

components/abstractions/src/main/java/com/microsoft/kiota/ApiExceptionBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+        return this;
+    }
+
+    /**
+     * Assign the response status code to the builder
+     * @param responseStatusCode an int representing the response status code.
+     * @return The builder object.
+     */
+    public ApiExceptionBuilder withResponseStatusCode(int responseStatusCode) {
```
baywet, line 62:
```suggestion
    @Nonnull
    public ApiExceptionBuilder withResponseStatusCode(int responseStatusCode) {
```

components/abstractions/src/main/java/com/microsoft/kiota/ApiExceptionBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+        value.setResponseStatusCode(responseStatusCode);
+        return this;
+    }
+
+    /**
+     * Assign the response headers to the builder
+     * @param responseHeaders the response headers to be added to this ApiException.
+     * @return The builder object.
+     */
+    public ApiExceptionBuilder withResponseHeaders(@Nonnull ResponseHeaders responseHeaders) {
```
baywet, line 75:
```suggestion
    @Nonnull
    public ApiExceptionBuilder withResponseHeaders(@Nonnull ResponseHeaders responseHeaders) {
```

components/abstractions/src/main/java/com/microsoft/kiota/ApiExceptionBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+        value.setResponseHeaders(responseHeaders);
+        return this;
+    }
+
+    /**
+     * Build and return an instance of ApiException
+     * @return The built ApiException.
+     */
+    public ApiException build() {
```
baywet, line 87:
```suggestion
    @Nonnull
    public ApiException build() {
```

baywet (changes requested):

baywet (approved):

PR #742, opened 2023-10-13
Preserve user errors instead of creating new ApiExceptions

andreaTP:
Bug found here: https://github.com/microsoft/kiota/pull/3401#discussion_r1356991436

There was too much to be done to test in this repo, I tested locally against the `kiota` integration tests.


baywet (approved):
Thanks for fixing this!

PR #766, opened 2023-10-30
Add a Basic Access Authentication Provider

andreaTP:
I have a few "clean-up" tasks from removing the hand-written Client in Apicurio Registry.
This is an easy one, but I'm looking for guidance on how to port the [Oidc one](https://github.com/Apicurio/apicurio-registry/blob/main/client/src/main/java/io/apicurio/registry/auth/OidcAccessTokenProvider.java) too.

The problem is that I need an HttpClient(or something like that / RequestAdapter etc.) to get the token(possibly during object creation), but this is used to construct the `RequestAdapter` object so we have a circular dependency.

Have you already faced this issue?
Do you have a solution that doesn't involve spinning a new, dedicated, HTTP client?

### DIFF
--- a/components/abstractions/src/main/java/com/microsoft/kiota/authentication/BasicAccessAuthenticationProvider.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/authentication/BasicAccessAuthenticationProvider.java
@@ -9,22 +9,28 @@ import java.util.Base64;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 
-// https://en.wikipedia.org/wiki/Basic_access_authentication
+/** Provides an implementation of the Basic Access Authentication scheme: https://en.wikipedia.org/wiki/Basic_access_authentication . */
 public class BasicAccessAuthenticationProvider implements AuthenticationProvider {
+    private final static String AUTHORIZATION_HEADER_KEY = "Authorization";
+    private static final String BASIC = "Basic ";
 
     private final String username;
     private final String password;
     private final String encoded;
 
-    public BasicAccessAuthenticationProvider(String username, String password) {
+    /**
+     * Instantiates a new BasicAccessAuthenticationProvider.
+     * @param username the username to be used.
+     * @param password the password to be used.
+     */
+    public BasicAccessAuthenticationProvider(@Nonnull final String username, @Nonnull final String password) {
         this.username = username;
         this.password = password;
         encoded = Base64.getEncoder().encodeToString((username + ":" + password).getBytes(StandardCharsets.UTF_8));
     }
 
-    private final static String AUTHORIZATION_HEADER_KEY = "Authorization";
-    public static final String BASIC = "Basic ";
     @Override
+    @Nonnull
     public CompletableFuture<Void> authenticateRequest(@Nonnull final RequestInformation request, @Nullable final Map<String, Object> additionalAuthenticationContext) {
         request.headers.add(AUTHORIZATION_HEADER_KEY, BASIC + encoded);
         return CompletableFuture.completedFuture(null);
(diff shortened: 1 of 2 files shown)


components/abstractions/src/main/java/com/microsoft/kiota/authentication/BasicAccessAuthenticationProvider.java
```diff
@@ -0,0 +1,32 @@
...
+import com.microsoft.kiota.RequestInformation;
+import jakarta.annotation.Nonnull;
+import jakarta.annotation.Nullable;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+// https://en.wikipedia.org/wiki/Basic_access_authentication
```
baywet, line 12: could you use a javadoc comment instead?

components/abstractions/src/main/java/com/microsoft/kiota/authentication/BasicAccessAuthenticationProvider.java
```diff
@@ -0,0 +1,32 @@
...
+public class BasicAccessAuthenticationProvider implements AuthenticationProvider {
+
+    private final String username;
+    private final String password;
+    private final String encoded;
+
+    public BasicAccessAuthenticationProvider(String username, String password) {
```
baywet, line 19:
```suggestion
    public BasicAccessAuthenticationProvider(@Nonnull final String username, @Nonnull final String password) {
```

components/abstractions/src/main/java/com/microsoft/kiota/authentication/BasicAccessAuthenticationProvider.java
```diff
@@ -0,0 +1,32 @@
...
+        this.username = username;
```
baywet, line 20: check for string being null of empty

baywet (changes requested):
Thanks for the contribution. I left a couple of comments.
Also if you want this published right away, maybe bump the minor?

andreaTP:
Fixed the spotBugs issues now.

> Also if you want this published right away, maybe bump the minor?

Not needed, good to have in the long run but I don't strictly need it.

More interested in feedback regarding the Oidc one 🙂

components/abstractions/src/main/java/com/microsoft/kiota/authentication/BasicAccessAuthenticationProvider.java
```diff
@@ -0,0 +1,32 @@
...
+        this.password = password;
+        encoded = Base64.getEncoder().encodeToString((username + ":" + password).getBytes(StandardCharsets.UTF_8));
+    }
+
+    private final static String AUTHORIZATION_HEADER_KEY = "Authorization";
+    public static final String BASIC = "Basic ";
+    @Override
+    public CompletableFuture<Void> authenticateRequest(@Nonnull final RequestInformation request, @Nullable final Map<String, Object> additionalAuthenticationContext) {
```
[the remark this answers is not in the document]
andreaTP: done

components/abstractions/src/main/java/com/microsoft/kiota/authentication/BasicAccessAuthenticationProvider.java
andreaTP: in other parts of the codebase we are only checking for null, done like that.

baywet (approved):

baywet:
As for the OIDC provider, I agree that having support for it would be super interesting.
From what I've read in the code:
- It depends on okHttp, so I don't think we could put it in abstractions, we might need to stand up another package.
- The caching concern is mixed with the token acquisition concern, I'd deport that to a dedicated interface + class so people can store the cached tokens wherever they want (thinking about multiple web servers with a backplane scenario)
- I'm not seeing code regarding the well known endpoints?

andreaTP:
Thanks a lot for getting back @baywet !

> - It depends on okHttp, so I don't think we could put it in abstractions

This is not a requirement, I can probably re-implement it based on the `RequestAdapter`

> - caching

I'm not 100% sure how it will look but we can work toward that, for sure

> - I'm not seeing code regarding the well known endpoints?

It's not used as the authentication mechanism that we are using depends only on a few variables that are passed through the constructor and they are not discovered dynamically.

PR #767, opened 2023-10-31
feature/serialization helpers

baywet:
related https://github.com/microsoft/kiota/issues/3406

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+## [0.8.0] - 2023-10-31
+
+### Added
+
 - Added a default implementation of `BasicAccessAuthenticationProvider`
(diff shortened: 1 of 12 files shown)


ramsessanchez:
@baywet , I'm wondering if the serializeAsString methods should be differentiated similar as to how you differentiate the deserialize methods. Perhaps serializeCollectionAsString or serializeCollectionAsStream? Otherwise looks good!

ramsessanchez (commented):
@baywet , I'm wondering if the serializeAsString methods should be differentiated similar as to how you differentiate the deserialize methods. Perhaps serializeCollectionAsString or serializeCollectionAsStream? Otherwise looks good!

baywet:
@ramsessanchez yeah I was wondering that too, we don't technically need to do it as the parameters types are different. It's a matter of consistency vs convenience. I believe the current situation feels familiar to the ecosystem (i.e. Java devs are used to overload with different parameters types)

PR #770, opened 2023-11-03
Revert from Async to Sync 

### LINKED ISSUE #175: [discussion] Java Async
I think it's worth having this conversation even just to keep track of the reasoning behind the decisions.
When I looked at the codebase, at first, I was pretty impressed by the fact that we are using async calls everywhere for Java.

This is a subject pretty much full of nuances in Java land as far as I know it.

Let's try to summarize a few key points:

 - `CompletableFuture` API is not great and doesn't play well with most of the rest of the standard library, also, using it makes the code much much harder to understand and follow (e.g. we miss an `async`/`await` construct)
 - [Reactive programming](https://www.reactivemanifesto.org/) has become quite popular in Java, and exposing a `CompletableFuture` interface is going to play nicely with all the frameworks adopting this paradigm (Akka, Vert-X etc. etc.)
 - Most of the libraries around are exposing still a blocking API(e.g. OkHttp), and, wrapping every call in `CompletableFuture`s might not be the ideal solution given the number of options that comes (scheduling on which Thread Pool? Keep separated IO blocking operations from in-process computing etc. etc.)
 - "Idiomatic" Java 8 code (as per Android compatibility) is not using the async API and it doesn't play nicely with imperative and blocking code
 - [Project Loom](https://wiki.openjdk.org/display/loom/Main) is coming with the promise to enable blocking idioms to easily work in contexts with concurrency and parallelism (since here we are keeping compatibility with Android I'm not sure about timelines here)

ramsessanchez:
Begin work to revert from Async to Sync for Java-Kiota components. 

fixes #175


components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
```diff
@@ -642,31 +635,22 @@ private Response throwIfFailedResponse(@Nonnull final Response response, @Nonnul
...
-            }};
+            final Map<String, Object> additionalContext = new HashMap<String, Object>();
+            additionalContext.put("parent-span", span);
             if(claims != null && !claims.isEmpty()) {
                 additionalContext.put(claimsKey, claims);
             }
-            return this.authProvider.authenticateRequest(requestInfo, additionalContext)
-            .thenCompose(x -> {
-                try {
-                    final OkHttpCallbackFutureWrapper wrapper = new OkHttpCallbackFutureWrapper();
```
baywet, line 659: The class OkHttpCallbackFutureWrapper should be deleted

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
```diff
@@ -642,31 +635,22 @@ private Response throwIfFailedResponse(@Nonnull final Response response, @Nonnul
...
-                    spanForAttributes.recordException(ex);
-                    final CompletableFuture<Response> result = new CompletableFuture<Response>();
-                    result.completeExceptionally(ex);
-                    return result;
-                }
-            })
-            .thenApply(x -> {
+            this.authProvider.authenticateRequest(requestInfo, additionalContext);
+            final OkHttpCallbackFutureWrapper wrapper = new OkHttpCallbackFutureWrapper();
+            this.client.newCall(getRequestFromRequestInformation(requestInfo, span, spanForAttributes)).enqueue(wrapper);
```
baywet, line 650: don't enqueue, just call. We don't want to be doing sync over async here.

baywet (changes requested):
Thanks for starting this, a couple of comments to help fasttrack it.

ramsessanchez:
appreciate the feedback and insight @baywet , thanks!

components/http/okHttp/src/main/java/com/microsoft/kiota/http/middleware/HeadersInspectionHandler.java
```diff
@@ -56,13 +57,17 @@ public Response intercept(final Chain chain) throws IOException {
             }
             if (inspectionOption.getInspectRequestHeaders()) {
                 for(final Pair<? extends String, ? extends String> header : request.headers()) {
-                    inspectionOption.getRequestHeaders().put(header.getFirst(), Set.of(header.getSecond()));
+                    HashSet<String> value = new HashSet<>();
```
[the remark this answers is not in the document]
ramsessanchez: Set.of() is a java 9+ feature

components/authentication/azure/src/main/java/com/microsoft/kiota/authentication/AzureIdentityAccessTokenProvider.java
```diff
@@ -80,15 +80,11 @@ public CompletableFuture<String> getAuthorizationToken(@Nonnull final URI uri, @
         try(final Scope scope = span.makeCurrent()) {
             if(!_hostValidator.isUrlHostValid(uri)) {
                 span.setAttribute("com.microsoft.kiota.authentication.is_url_valid", false);
-                return CompletableFuture.completedFuture("");
+                return "";
             }
             if(!uri.getScheme().equalsIgnoreCase("https")) {
                 span.setAttribute("com.microsoft.kiota.authentication.is_url_valid", false);
-                final Exception result = new IllegalArgumentException("Only https is supported");
-                span.recordException(result);
```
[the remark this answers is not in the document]
ramsessanchez: if we throw it here then it gets caught at the catch phase and recorded there, I was following this general pattern in cases where there were multiple types of exceptions thrown to only have a single span.recordException() call.

components/abstractions/src/main/java/com/microsoft/kiota/ResponseHandler.java
```diff
@@ -13,12 +12,12 @@
...
+     * @param response             The native response object.
+     * @param errorMappings        the error mappings for the response to use when deserializing failed responses bodies. Where an error code like 401 applies specifically to that status code, a class code like 4XX applies to all status codes within the range if an the specific error code is not present.
      * @param <NativeResponseType> The type of the native response object.
-     * @param <ModelType> The type of the response model object.
-     * @return A CompletableFuture that represents the asynchronous operation and contains the deserialized response.
+     * @param <ModelType>          The type of the response model object.
      */
     @Nonnull
-    <NativeResponseType, ModelType> CompletableFuture<ModelType> handleResponseAsync(@Nonnull final NativeResponseType response, @Nullable final HashMap<String, ParsableFactory<? extends Parsable>> errorMappings);
+    <NativeResponseType, ModelType> void handleResponseAsync(@Nonnull final NativeResponseType response, @Nullable final HashMap<String, ParsableFactory<? extends Parsable>> errorMappings);
```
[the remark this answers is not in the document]
ramsessanchez: yep that makes sense, not sure what I was thinking here

components/abstractions/src/main/java/com/microsoft/kiota/RequestAdapter.java
```diff
@@ -35,7 +38,7 @@ public interface RequestAdapter {
      */
     @Nullable
     @SuppressWarnings("LambdaLast")
-    <ModelType extends Parsable> CompletableFuture<ModelType> sendAsync(@Nonnull final RequestInformation requestInfo, @Nonnull final ParsableFactory<ModelType> factory, @Nullable final HashMap<String, ParsableFactory<? extends Parsable>> errorMappings);
+    <ModelType extends Parsable> ModelType sendAsync(@Nonnull final RequestInformation requestInfo, @Nonnull final ParsableFactory<ModelType> factory, @Nullable final HashMap<String, ParsableFactory<? extends Parsable>> errorMappings) throws IOException, URISyntaxException;
```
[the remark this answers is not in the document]
ramsessanchez: Gotcha, I think exception catching/throwing was my main concern regarding this work, so I appreciate the clarity here.

components/authentication/azure/src/main/java/com/microsoft/kiota/authentication/AzureIdentityAccessTokenProvider.java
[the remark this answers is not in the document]
baywet: ah ok didn't see that section of the code. Resolving.

components/http/okHttp/src/main/java/com/microsoft/kiota/http/middleware/HeadersInspectionHandler.java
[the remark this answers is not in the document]
baywet: Thanks for clarifying. I'm not sure why the CI didn't complain about that earlier...

components/abstractions/src/main/java/com/microsoft/kiota/authentication/ApiKeyAuthenticationProvider.java
```diff
@@ -64,14 +62,11 @@ public CompletableFuture<Void> authenticateRequest(@Nonnull final RequestInforma
             final URI uri = request.getUri();
             if(uri == null || !validator.isUrlHostValid(uri)) {
                 span.setAttribute("com.microsoft.kiota.authentication.is_url_valid", false);
-                return CompletableFuture.completedFuture(null);
+                return;
             }
             if(!uri.getScheme().equalsIgnoreCase("https")) {
                 span.setAttribute("com.microsoft.kiota.authentication.is_url_valid", false);
-                final Exception result = new IllegalArgumentException("Only https is supported");
-                span.recordException(result);
```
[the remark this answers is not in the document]
ramsessanchez: recording it at the catch block further down.

components/abstractions/src/main/java/com/microsoft/kiota/serialization/SerializationWriterProxyFactory.java
```diff
@@ -24,7 +24,7 @@ public String getValidContentType() {
      * @param onAfterSerialization the callback to invoke after the serialization of any model object.
      * @param onStartObjectSerialization the callback to invoke when the serialization of a model object starts.
      */
-    public SerializationWriterProxyFactory(@Nonnull final SerializationWriterFactory concrete,
```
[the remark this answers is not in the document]
ramsessanchez: The usages for this were all internal, wasn't sure about the scenario where this would be publicly used.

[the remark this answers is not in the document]
baywet: consumers might use it while configuring the request adapter when they don't want to rely on the default registry

components/authentication/azure/src/main/java/com/microsoft/kiota/authentication/AzureIdentityAccessTokenProvider.java
```diff
@@ -113,8 +106,13 @@ public CompletableFuture<String> getAuthorizationToken(@Nonnull final URI uri, @
             if(decodedClaim != null && !decodedClaim.isEmpty()) {
                 context.setClaims(decodedClaim);
             }
-            return this.creds.getToken(context).toFuture().thenApply(r -> r.getToken());
-        } finally {
+            return this.creds.getToken(context).toFuture().join().getToken();
```
andrueastman, line 109: Any reason to not use the sync overload?
```suggestion
            return this.creds.getTokenSync(context).getToken();
```

components/abstractions/src/test/java/com/microsoft/kiota/authentication/ApiKeyAuthenticationProviderTest.java
```diff
@@ -20,34 +20,32 @@ public void DefensivePrograming() {
...
-	public void AddsInQueryParameterWithOtherParameters() throws IllegalStateException, URISyntaxException {
+	void AddsInQueryParameterWithOtherParameters() throws IllegalStateException, URISyntaxException {
 		var value = new ApiKeyAuthenticationProvider("key", "param", ApiKeyLocation.QUERY_PARAMETER);
 		var request = new RequestInformation() {{
 			urlTemplate = "https://localhost{?param1}";
 		}};
 		request.addQueryParameter("param1", "value1");
-		value.authenticateRequest(request, null).join();
-		assertNull(request.headers.get("param"));
+		value.authenticateRequest(request, null);
```
andrueastman, line 39: No reason to drop the assert
```suggestion
		value.authenticateRequest(request, null);
		assertNull(request.headers.get("param"));
```

components/abstractions/src/main/java/com/microsoft/kiota/authentication/BaseBearerTokenAuthenticationProvider.java
```diff
@@ -37,21 +35,13 @@ public CompletableFuture<Void> authenticateRequest(@Nonnull final RequestInforma
...
                 targetUri = request.getUri();
-            } catch (URISyntaxException e) {
-                final CompletableFuture<Void> result = new CompletableFuture<>();
-                result.completeExceptionally(e);
-                return result;
+            } catch (URISyntaxException e){
+                throw new RuntimeException("Malformed URI.", e);
+            }
+            String accessToken = this.accessTokenProvider.getAuthorizationToken(targetUri, additionalAuthenticationContext);
+            if(!Strings.isNullOrEmpty(accessToken)) {
```
andrueastman, line 42:
```suggestion
            if(!Compatibility.isBlank(accessToken)) {
```

components/abstractions/src/main/java/com/microsoft/kiota/MultipartBody.java
```diff
@@ -2,11 +2,7 @@
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.HashMap;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Objects;
-import java.util.UUID;
+import java.util.*;
```
andrueastman, line 5: The `import com.google.common.base.Strings;` on line 11 is now unused.

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
```diff
@@ -306,243 +291,222 @@ private String getMediaTypeAndSubType(@Nonnull final MediaType mediaType) {
...
+                    closeResponse(closeResponse, response);
                 }
-            });
+            } else {
+                span.addEvent(eventResponseHandlerInvokedKey);
+                return responseHandler.handleResponse(response, errorMappings);
+            }
+
+
+        } catch(ApiException ex) {
```
baywet, line 463: @andreaTP I'd like your opinion here. All generated exceptions derive from APIexception, which is checked.
Do you think we should:
- wrap the exception in a runtime one like we're doing here? (least favorite)
- declare the exception on the API surface (for the request adapter interface, the implementation, and the request executors), which would lead to consumers having to handle that exception?
- change the exception to be unchecked?

andreaTP: Thanks a lot for sharing @baywet !

On this subject, I do have a strong opinion, but it might or might not be aligned with the rest of the Java community.
I would:

> change the exception to be unchecked

Among the top motivations:

- Checked Exceptions do not play well with lambdas, functional interface, and most of the more "modern" std lib API constructs
- Checked Exceptions will cause major disruption when an OpenAPI changes (e.g. if you declare an additional error mapping you have to propagate it in all of the generated and the code using it)

I can provide more evidence if needed/required, but let's try to get @EricWittmann feedback here as well.

baywet: to your second point, since all mapped errors derive from Api Exception, wouldn't declaring ApiException be enough in the case we stick with checked exception?

andreaTP: correct, that's possible, but it depends on the user code.

components/authentication/azure/src/main/java/com/microsoft/kiota/authentication/AzureIdentityAccessTokenProvider.java
ramsessanchez: Good catch

components/http/okHttp/gradle/dependencies.gradle
```diff
@@ -11,7 +11,7 @@ dependencies {
     // This dependency is used internally, and not exposed to consumers on their own compile classpath.
     implementation 'io.opentelemetry:opentelemetry-api:1.31.0'
     implementation 'io.opentelemetry:opentelemetry-context:1.31.0'
-    implementation 'io.opentelemetry:opentelemetry-semconv:1.30.1-alpha'
+    implementation 'io.opentelemetry.semconv:opentelemetry-semconv:1.22.0-alpha'
```
baywet, line 14: can you provide the rationale here? it feels like a step back version wise

ramsessanchez:
```
Deprecated
Use io.opentelemetry.semconv.SemanticAttributes from io.opentelemetry.semconv:opentelemetry-semconv:{{version}}  instead. 
```
The previous one was the old dependency path and the second is the new one, it seems this one is fully dedicated to the semconv library. 1.22.0 is the latest for this version.

baywet: ok thanks for the additional context here

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
EricWittmann: For my money it's unchecked exceptions all the way for code like this.  So much easier to work with, and you can still catch whatever exceptions you want.  It might make it **slightly** less explicit but so much easier to work with.  Just my $0.02.  :)

baywet: alright, thanks for the input here, I think we'll rely on your domain knowledge on this one.
@ramsessanchez can you:
- update ApiException to derive from an unchecked exception instead.
- update the code to stop wrapping up ApiException (or derived from generated code) in runtime exception.
- remove any throws ApiException from the method declarations.
Please?

andrueastman (dismissed):

CHANGELOG.md
```diff
@@ -10,7 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-Changed OkHttpRequestAdapter dependency from OkHttpClient to Call.Factory (parent interface implemented by OkHttpClient).
+## [1.0.0] - 2023-11-10
```
baywet, line 13: can we use a minor version for the time being? And want to make sure things are aligned before we major rev.

components/http/okHttp/src/main/java/com/microsoft/kiota/http/HeadersCompatibility.java
```diff
@@ -0,0 +1,49 @@
...
+import com.microsoft.kiota.ResponseHeaders;
+import jakarta.annotation.Nonnull;
+
+import java.util.HashSet;
+import java.util.Objects;
+
+/**
+ * Compatibility class to bridge OkHttp Headers and Kiota Headers
+ */
+public class HeadersCompatibility {
```
baywet, line 13: could we scope it to package instead of public?

gradle.properties
```diff
@@ -24,8 +24,8 @@ org.gradle.parallel=true
 org.gradle.caching=true
 
 mavenGroupId         = com.microsoft.kiota
-mavenMajorVersion = 0
```
baywet, line 27: same comment about using a minor version instead

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
```diff
@@ -642,45 +566,35 @@ private Response throwIfFailedResponse(@Nonnull final Response response, @Nonnul
...
-                if(contentTypeHeaderValue != null && !contentTypeHeaderValue.isEmpty()) {
-                    spanForAttributes.setAttribute("http.response_content_type", contentTypeHeaderValue);
-                }
-                spanForAttributes.setAttribute(SemanticAttributes.HTTP_STATUS_CODE, x.code());
-                spanForAttributes.setAttribute(SemanticAttributes.HTTP_FLAVOR, x.protocol().toString().toUpperCase(Locale.ROOT));
-                return x;
-            })
-            .thenCompose(x -> this.retryCAEResponseIfRequired(x, requestInfo, span, spanForAttributes, claims));
+            this.authProvider.authenticateRequest(requestInfo, additionalContext);
+            this.client.newCall(getRequestFromRequestInformation(requestInfo, span, spanForAttributes));
```
baywet, line 580:
```suggestion
```
I'm not sure why we're making the call twice here?

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
```diff
@@ -642,45 +566,35 @@ private Response throwIfFailedResponse(@Nonnull final Response response, @Nonnul
...
+            Response response = this.client.newCall(getRequestFromRequestInformation(requestInfo, span, spanForAttributes)).execute();
```
baywet, line 581:
```suggestion
            final Response response = this.client.newCall(getRequestFromRequestInformation(requestInfo, span, spanForAttributes)).execute();
```

baywet (changes requested):
Getting really close!

components/http/okHttp/src/main/java/com/microsoft/kiota/http/HeadersCompatibility.java
ramsessanchez: It's used in core under a different namespace, so had to make it public.

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
ramsessanchez: oversight I believe, will correct in the next commit.

components/http/okHttp/src/main/java/com/microsoft/kiota/http/HeadersCompatibility.java
baywet: but core, do you mean msgraph core or something else?

ramsessanchez: In java-core. I needed to translate the headers from okhttp->kiota headers when using the ApiException builder and so I added this class so I could use your re-use your logic and also perhaps if needed again in the future.

ramsessanchez: https://github.com/microsoftgraph/msgraph-sdk-java-core/blob/a4857861f70872716f5f63d7960f622932ed8c6a/src/main/java/com/microsoft/graph/requests/ResponseBodyHandler.java#L80

baywet: ah ok, thanks for the additional context, resolving

baywet (approved):
thanks for making the changes

PR #804, opened 2023-11-08
feature/code reduction generators

baywet:
related https://github.com/microsoft/kiota/issues/3651


components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
```diff
@@ -38,6 +38,46 @@ public class RequestInformation {
...
+        configure(requestConfiguration, configurationFactory, null);
+    }
+    /**
+     * Configures the request information based on the request configuration and the query parameters getter.
+     * @param <T> The type of the request configuration.
+     * @param requestConfiguration The request configuration to apply to the request information.
+     * @param configurationFactory The factory to create the request configuration from.
+     * @param queryParametersGetter The function to get the query parameters from the request configuration.
+     */
+    public <T extends BaseRequestConfiguration> void configure(@Nullable final java.util.function.Consumer<T> requestConfiguration, @Nonnull final java.util.function.Supplier<T> configurationFactory, @Nullable final java.util.function.Function<T, Object> queryParametersGetter) {
```
baywet, line 68: @andreaTP right now the code being generated looks like this

```java
final RequestInformation requestInfo = new RequestInformation(HttpMethod.GET, urlTemplate, pathParameters);
        requestInfo.configure(requestConfiguration, GetRequestConfiguration::new, x -> x.queryParameters);
        requestInfo.headers.tryAdd("Accept", "application/json;q=1");
        return requestInfo;
```

The reason why we're providing a callback to get the query parameter is because the request configuration class is still being projected on its own. (e.g. GetRequestConfiguration here).
And that's the case because I don't want to use reflection to create a new instance of the query parameters in the BaseRequestConfiguration.
But if we moved the query parameters property to the base class, made it generic, we could instead have that, and do away with the GetRequestConfiguration class all together.
Which would result in 
```java
final RequestInformation requestInfo = new RequestInformation(HttpMethod.GET, urlTemplate, pathParameters);
        requestInfo.configure(requestConfiguration, () -> {
              final RequestConfiguration<GetQueryParameters> config = new RequestConfiguration<GetQueryParameters>();
              config.queryParameters = new GetQueryParameters();
              return config;
        });
        requestInfo.headers.tryAdd("Accept", "application/json;q=1");
        return requestInfo;
```

Thoughts?

andreaTP: Sorry for the delay, a little sick those days ...
I think I do not get the "aim" of this PR, the signature you are mentioning seems to be good to me, but the latter example is growing the code necessary to invoke it.

I might be missing something.

baywet: No worries, I hope you get better soon.

The goal here is to reduce the amount of code we generate in the request generator bodies. (toXXXRequestInformation)

While working on this, I was pondering whether we could "hoist" the query parameter property as a generic property on the request configuration class in abstractions, the goal being to avoid passing a lambda to the configure method and to remove the generated request configuration class all together.

But if we do so, we still need to instantiate the query parameters generic property. Because of type erasure, the only way I know how to do this is through this other complex lambda, which feels like defeating the point.

andreaTP: > But if we do so, we still need to instantiate the query parameters generic property. Because of type erasure, the only way I know how to do this is through this other complex lambda, which feels like defeating the point.

Agree, that's why I was not getting it probably.

I haven't thought this entirely through, but an alternative workflow that might spare a bunch of instructions is to get completely rid of lambdas here and pass along the chain a pre-configured config object that the user will mutate in place. Apart from the code-size benefits this will probably be also much faster. But I might be missing some pieces ... I would need to try ...

baywet: yes that's the other alternative, instead of asking the developer for a `Consumer<RequestOption<QueryParameters>>` we could directly ask them for `RequestOption<QueryParameters>`. The main reason why we originally went with lamdbas was because we wanted to avoid requiring the consumer to new up the thing.
But in Go for example, because this doesn't feel idiomatic, we ask the instance directly (later case).
Do you have an opinion from a developer experience perspective?

andreaTP: If you don't want the user to call "new" you can simply pass the instance after creating it (and possibly applying the default values).
The user will be able to modify the configuration as a side effect and you are going to save a few allocations too.
I think this is very much turning in my original proposal [here](https://github.com/microsoft/kiota/issues/3378), and I'm super happy to see this happening 🙂

baywet: right, but with this proposal:
- any call to the executor method that needs configuration, will require two calls to the fluent API instead of one.
- we'll loose the strongly typed query parameters.

andreaTP: > any call to the executor method that needs configuration, will require two calls to the fluent API instead of one.

I might be missing something, but no, the API can remain pretty much the same.

> we'll loose the strongly typed query parameters.

why?

baywet: ok I probably didn't understand your initial suggestion. Can you please reiterate here how one would configure a query parameter on the request executor?

andreaTP: I'll try to come up with a POC example next week if it's not too late.

baywet: deferred this discussion here as this experience point is actually orthogonal.
https://github.com/microsoft/kiota-java/issues/806

ramsessanchez (approved):

PR #805, opened 2023-11-08
Use parent Call.Factory of OkHttpClient

cbos:
@baywet 
Same as https://github.com/microsoftgraph/msgraph-sdk-java-core/pull/1252

Call.Factory is the parent interface of OkHttpClient and allows adding your own OkHttpClient (wrapper) to be used, which can instrument the Http requests.


baywet (approved):
Thank you for porting the change!

PR #822, opened 2023-11-16
Review transitive dependencies

andreaTP:
I reviewed the transitive dependencies, realized that some have been used for very little value, and switched to "standard library" based alternatives.


baywet (approved):
Thank you for cleaning this up! It's going to make things even leaner!!
@ramsessanchez we should look at how much we need guava in msgraph core

PR #823, opened 2023-11-16
[chore] Apply spotless

andreaTP:
As we are going for GA and, after the big changes for sync/async the GH history is already a bit compromised, I think this is the right time to introduce this change.

How to review:

- the spotless configuration is at this commit: https://github.com/microsoft/kiota-java/commit/c095aefa465a0e1ecfbb6a97af35449c35345867
- feel free to restart from there and add any different configuration you might like and `./gradlew spotlessApply` to see how it looks


components/abstractions/src/main/java/com/microsoft/kiota/BaseRequestBuilder.java
```diff
@@ -1,47 +1,55 @@
...
 import java.util.Objects;
 
-import jakarta.annotation.Nonnull;
-
 /** Base class for request builders */
 public abstract class BaseRequestBuilder {
     /** Path parameters for the request */
-    @Nonnull
-    protected HashMap<String, Object> pathParameters;
+    @Nonnull protected HashMap<String, Object> pathParameters;
```
baywet, line 10: is there a way to configure the behaviour? I'm not sure this kind of change is super helpful

components/abstractions/src/main/java/com/microsoft/kiota/RequestAdapter.java
```diff
@@ -32,9 +32,12 @@ public interface RequestAdapter {
      * @param <ModelType> the type of the response model to deserialize the response into.
      * @return the deserialized response model.
      */
-    @Nullable
-    @SuppressWarnings("LambdaLast")
-    <ModelType extends Parsable> ModelType send(@Nonnull final RequestInformation requestInfo, @Nonnull final ParsableFactory<ModelType> factory, @Nullable final HashMap<String, ParsableFactory<? extends Parsable>> errorMappings);
+    @Nullable @SuppressWarnings("LambdaLast")
```
baywet, line 35: same thing here, I'm not a big fan of that change either. But I might be biased by dotnet.

baywet (commented):
Thanks for setting this up. It's a great improvement to our dev loop.
I did have a couple of questions.
Also, do you think it'd be possible to enforce things are properly formatted as part of the build/CI process?

components/abstractions/src/main/java/com/microsoft/kiota/RequestAdapter.java
andreaTP: No strong opinions here, I let you take the lead, I'm happy with whatever format as long as it's automated.

andreaTP:
> Also, do you think it'd be possible to enforce things are properly formatted as part of the build/CI process?

This is already done, `./gradlew build` will fail if files are not formatted.

build/tmp/spotless-register-dependencies
```diff
@@ -0,0 +1 @@
+1
```
baywet, line 1: should this file be here? or should we add it to the git ignore?

components/abstractions/src/main/java/com/microsoft/kiota/BaseRequestBuilder.java
baywet: Actually, after [reading the documentation](https://github.com/diffplug/spotless/tree/main/plugin-gradle#formatannotations) I retract my objection here. This annotation qualifies the Type of the field, this is why it's inline. An annotation qualifying the field itself would be at the line above. It's different from dotnet, but the rationale makes sense. I'll just have to get used to it. Thanks for sharing the context.

components/abstractions/src/main/java/com/microsoft/kiota/RequestAdapter.java
baywet: (fixes the configuration for SuppressWarnings, resolving)

build/tmp/spotless-register-dependencies
baywet: after some local testing, it appears this file is generated on build and can be removed safely.

baywet (approved):

andreaTP:
Amazing! Thanks a ton for making this happen @baywet !!! ❤️

PR #839, opened 2023-11-20
Json parse improvements

andreaTP:


components/serialization/json/src/main/java/com/microsoft/kiota/serialization/JsonParseNode.java
```diff
@@ -348,15 +271,10 @@ private <T extends Parsable> void assignFieldValues(
...
-                    final Consumer<Parsable> onBefore = this.onBeforeAssignFieldValues;
-                    final Consumer<Parsable> onAfter = this.onAfterAssignFieldValues;
-                    fieldDeserializer.accept(
-                            new JsonParseNode(fieldValue) {
-                                {
-                                    this.setOnBeforeAssignFieldValues(onBefore);
-                                    this.setOnAfterAssignFieldValues(onAfter);
-                                }
-                            });
+                    JsonParseNode itemNode = new JsonParseNode(fieldValue);
```
baywet, line 274:
```suggestion
                    final JsonParseNode itemNode = new JsonParseNode(fieldValue);
```

components/serialization/json/src/main/java/com/microsoft/kiota/serialization/JsonParseNode.java
```diff
@@ -248,40 +198,13 @@ public T next() {
...
-                                }
-                            };
-                        }
-                    };
-
-            for (T elem : iterable) {
-                result.add(elem);
+            while (sourceIterator.hasNext()) {
+                JsonElement item = sourceIterator.next();
+                JsonParseNode itemNode = new JsonParseNode(item);
```
baywet, line 204:
```suggestion
                final JsonElement item = sourceIterator.next();
                final JsonParseNode itemNode = new JsonParseNode(item);
```

components/serialization/json/src/main/java/com/microsoft/kiota/serialization/JsonParseNode.java
```diff
@@ -203,38 +179,12 @@ public T next() {
...
```
baywet, line 184:
```suggestion
                final JsonElement item = sourceIterator.next();
                final JsonParseNode itemNode = new JsonParseNode(item);
```

components/serialization/json/src/main/java/com/microsoft/kiota/serialization/JsonParseNode.java
```diff
@@ -125,69 +158,12 @@ public JsonParseNode(@Nonnull final JsonElement node) {
...
```
baywet, line 163:
```suggestion
                final JsonElement item = sourceIterator.next();
                final JsonParseNode itemNode = new JsonParseNode(item);
```

components/serialization/json/src/main/java/com/microsoft/kiota/serialization/JsonParseNode.java
```diff
@@ -125,69 +158,12 @@ public JsonParseNode(@Nonnull final JsonElement node) {
...
-                                    }
-                                }
-                            };
-                        }
-                    };
-
-            for (T elem : iterable) {
-                result.add(elem);
+            while (sourceIterator.hasNext()) {
+                JsonElement item = sourceIterator.next();
```
baywet, line 162: we have those 4 first lines at 4 different places, do you think we could use a method to get the JsonParseNode from a JsonElement and set the the before/after assign field values callbacks?

baywet (changes requested):
Thanks for cleaning this up, a couple of suggestions here.

components/serialization/json/src/main/java/com/microsoft/kiota/serialization/JsonParseNode.java
andreaTP: I accept those suggestions, but I appreciate that this is a style coming strongly from dotnet.
In java codebases, I have not seen much declaring `final` in function bodies, it doesn't hurt though.

components/serialization/json/src/main/java/com/microsoft/kiota/serialization/JsonParseNode.java
andreaTP: no strong opinion here, I don't know how much those routines are expected to change and, possibly, diverge over time.

baywet: I'd say not at all, besides the last line which is different for each block

andreaTP: ok, let me refactor this too

andreaTP: refactored the iteration on the Array, I noticed only after that's slight different from what you asked for, but I think it reads pretty well now ... let me know

baywet (approved):

PR #843, opened 2023-11-21
Remove enum reflection

### LINKED ISSUE #840: Avoid usage of reflection API in Enum deserialization
https://github.com/microsoft/kiota-java/blob/033afd138a5692a7baf1f2003b4669476a5f7110/components/serialization/json/src/main/java/com/microsoft/kiota/serialization/JsonParseNode.java#L309

This is extremely unsafe and it works only because there is a "hidden" contract with the generator to emit the specific `forValue` method on the enum class.
My first reaction would be to make it more similar to a normal "class", and extend the API of `ValuedEnum` to cover the deserialization explicitly.

Please note that this is going to be a breaking API change for abstractions and I hope we can fix it before GA.

Do you have any prior art on the subject?
Any consideration?

andreaTP:
Proposal to fix #840

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+## [0.9.2] - 2023-11-16
+
 ### Changed
 
 - Added Spotless as an automatic formatting tool for the entire codebase
 - Changed some internal implementations of JsonParse for performance and readability reasons
-- [breaking] Removed the usage of reflection for enum deserialization
+- [breaking] Removed the usage of reflection for enum deserialization and reordered `RequestAdapter` arguments order
 
 ## [0.9.2] - 2023-11-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added Spotless as an automatic formatting tool for the entire codebase
 - Changed some internal implementations of JsonParse for performance and readability reasons
+- [breaking] Removed the usage of reflection for enum deserialization
 
 ## [0.9.2] - 2023-11-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-## [0.9.2] - 2023-11-16
+## [0.10.0] - 2023-11-22
 
 ### Changed
 
--- a/components/abstractions/src/main/java/com/microsoft/kiota/RequestAdapter.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/RequestAdapter.java
@@ -8,6 +8,7 @@ import jakarta.annotation.Nonnull;
 import jakarta.annotation.Nullable;
 import java.util.HashMap;
 import java.util.List;
+import java.util.function.Function;
 
 /** Service responsible for translating abstract Request Info into concrete native HTTP requests. */
 public interface RequestAdapter {
@@ -80,27 +81,27 @@ public interface RequestAdapter {
     /**
      * Executes the HTTP request specified by the given RequestInformation and returns the deserialized enum value.
      * @param requestInfo the request info to execute.
-     * @param targetClass the class of the response model to deserialize the response into.
+     * @param forValue a function that retrieves an enum from the corresponding string.
      * @param errorMappings the error factories mapping to use in case of a failed request.
      * @param <ModelType> the type of the response model to deserialize the response into.
      * @return the deserialized primitive response model.
      */
     @Nullable <ModelType extends Enum<ModelType>> ModelType sendEnum(
             @Nonnull final RequestInformation requestInfo,
-            @Nonnull final Class<ModelType> targetClass,
+            @Nonnull final Function<String, ModelType> forValue,
             @Nullable final HashMap<String, ParsableFactory<? extends Parsable>> errorMappings);
 
     /**
      * Executes the HTTP request specified by the given RequestInformation and returns the deserialized enum collection value.
      * @param requestInfo the request info to execute.
-     * @param targetClass the class of the response model to deserialize the response into.
+     * @param forValue a function that retrieves an enum from the corresponding string.
      * @param errorMappings the error factories mapping to use in case of a failed request.
      * @param <ModelType> the type of the response model to deserialize the response into.
      * @return the deserialized primitive response model.
      */
     @Nullable <ModelType extends Enum<ModelType>> List<ModelType> sendEnumCollection(
             @Nonnull final RequestInformation requestInfo,
-            @Nonnull final Class<ModelType> targetClass,
+            @Nonnull final Function<String, ModelType> forValue,
             @Nullable final HashMap<String, ParsableFactory<? extends Parsable>> errorMappings);
 
     /**
--- a/components/abstractions/src/main/java/com/microsoft/kiota/RequestAdapter.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/RequestAdapter.java
@@ -32,7 +32,6 @@ public interface RequestAdapter {
      * @param <ModelType> the type of the response model to deserialize the response into.
      * @return the deserialized response model.
      */
-    @SuppressWarnings("LambdaLast")
     @Nullable <ModelType extends Parsable> ModelType send(
             @Nonnull final RequestInformation requestInfo,
             @Nullable final HashMap<String, ParsableFactory<? extends Parsable>> errorMappings,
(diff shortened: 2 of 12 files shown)


components/abstractions/src/main/java/com/microsoft/kiota/serialization/ParseNode.java
```diff
@@ -110,18 +111,18 @@ public interface ParseNode {
...
 
     /**
      * Gets the EnumSet value of the node.
      * @return the EnumSet value of the node.
-     * @param targetEnum the class of the enum.
+     * @param forValue a function that given a string returns the corresponding enum
      * @param <T> the type of the enum.
      */
-    @Nullable <T extends Enum<T>> EnumSet<T> getEnumSetValue(@Nonnull final Class<T> targetEnum);
+    @Nullable <T extends Enum<T>> EnumSet<T> getEnumSetValue(@Nonnull final Function<String, T> forValue);
```
andreaTP, line 125: I haven't found any usage of this method anywhere in the codebase nor in the generator repository.
`EnumSet` does have a very idiosyncratic API and the semantics of the method looks similar to `getCollectionOfEnumValues` AFAIU.
Given the [Occam's Razor](https://en.wikipedia.org/wiki/Occam%27s_razor) principle, should we consider removing it at all?

baywet: The reason why enum sets are in the code base is to support "flaggable" enums. But there are limitations currently in the OAS spec and in the generation in tother languages, so we tabled it for a later major version.
see https://github.com/microsoft/kiota/issues/3238 and linked issue for the details.
We could leave it here for now, I don't think it hurts anyone even though it's not used today. If we add it back after GA, it'll be a breaking change with all the ceremony that's required, and our plans are not firm yet.

baywet (approved):
Thanks for taking this on. I think we should bump the minor and release the packages as soon as the generation changes are in.

components/abstractions/src/main/java/com/microsoft/kiota/serialization/ParseNode.java
andreaTP: Thanks for the explanation, I'm ok leaving it here, I doubt that `EnumSet` is going to be the final API, but we will see.

components/abstractions/src/main/java/com/microsoft/kiota/RequestAdapter.java
```diff
@@ -80,27 +81,27 @@ public interface RequestAdapter {
     /**
      * Executes the HTTP request specified by the given RequestInformation and returns the deserialized enum value.
      * @param requestInfo the request info to execute.
-     * @param targetClass the class of the response model to deserialize the response into.
+     * @param forValue a function that retrieves an enum from the corresponding string.
      * @param errorMappings the error factories mapping to use in case of a failed request.
      * @param <ModelType> the type of the response model to deserialize the response into.
      * @return the deserialized primitive response model.
      */
     @Nullable <ModelType extends Enum<ModelType>> ModelType sendEnum(
```
baywet, line 89:
```suggestion
    @SuppressWarnings("LambdaLast")
    @Nullable <ModelType extends Enum<ModelType>> ModelType sendEnum(
```

components/abstractions/src/main/java/com/microsoft/kiota/RequestAdapter.java
```diff
@@ -80,27 +81,27 @@ public interface RequestAdapter {
...
     /**
      * Executes the HTTP request specified by the given RequestInformation and returns the deserialized enum collection value.
      * @param requestInfo the request info to execute.
-     * @param targetClass the class of the response model to deserialize the response into.
+     * @param forValue a function that retrieves an enum from the corresponding string.
      * @param errorMappings the error factories mapping to use in case of a failed request.
      * @param <ModelType> the type of the response model to deserialize the response into.
      * @return the deserialized primitive response model.
      */
     @Nullable <ModelType extends Enum<ModelType>> List<ModelType> sendEnumCollection(
```
baywet, line 102:
```suggestion
    @SuppressWarnings("LambdaLast")
    @Nullable <ModelType extends Enum<ModelType>> List<ModelType> sendEnumCollection(
```

baywet (changes requested):

andreaTP:
I'm converting from Lambdas to plain classes to achieve compatibility, leave this on me @baywet if you don't mind.

baywet:
> I'm converting from Lambdas to plain classes to achieve compatibility, leave this on me @baywet if you don't mind.

do you mean using this syntax instead of a lambda `EnumName::forValue` ?

andreaTP:
I done the changes to make the linter happy and expanded the functional interface into a real `interface`, let me know if this looks good.

components/abstractions/src/main/java/com/microsoft/kiota/serialization/ValuedEnumParser.java
```diff
@@ -0,0 +1,13 @@
+package com.microsoft.kiota.serialization;
+
+import jakarta.annotation.Nonnull;
+
+/** The interface for a valued enum parser. */
+public interface ValuedEnumParser<T extends Enum> {
```
baywet, line 6:
```suggestion
@FunctionalInterface
public interface ValuedEnumParser<T extends Enum> {
```

components/abstractions/src/main/java/com/microsoft/kiota/RequestAdapter.java
```diff
@@ -80,28 +81,28 @@ public interface RequestAdapter {
...
      * @param errorMappings the error factories mapping to use in case of a failed request.
+     * @param enumParser a parser from string to enum instances.
      * @param <ModelType> the type of the response model to deserialize the response into.
      * @return the deserialized primitive response model.
      */
     @Nullable <ModelType extends Enum<ModelType>> ModelType sendEnum(
             @Nonnull final RequestInformation requestInfo,
-            @Nonnull final Class<ModelType> targetClass,
-            @Nullable final HashMap<String, ParsableFactory<? extends Parsable>> errorMappings);
+            @Nullable final HashMap<String, ParsableFactory<? extends Parsable>> errorMappings,
```
baywet, line 91: I'm not sure about the ordering change, this is going to be inconsistent with the object methods?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added Spotless as an automatic formatting tool for the entire codebase
```
baywet, line 13: changelog now needs to be updated

baywet (changes requested):

components/abstractions/src/main/java/com/microsoft/kiota/RequestAdapter.java
andreaTP: The ordering change has been done due to the linter complaining about Kotlin compatibility.
It's not consistent at the moment, should I make it consistent in this way:
- requestInfo
- errorMappings
- additionalthings
?

baywet: sure, let's take this opportunity to align things and clean the lambda last suppressions, thanks for calling this out

andreaTP: done

andreaTP:
maybe not, let me double-check, sorry for the noise.

components/serialization/json/src/test/java/com/microsoft/kiota/serialization/mocks/TestEntity.java
```diff
@@ -131,7 +131,12 @@ public Map<String, Consumer<ParseNode>> getFieldDeserializers() {
                 put(
                         "myEnum",
                         (n) -> {
-                            setMyEnum(n.getEnumValue(MyEnum.class));
+                            setMyEnum(
+                                    n.getEnumValue(
+                                            value ->
+                                                    (value == null || MyEnum.values().length < 1)
+                                                            ? null
+                                                            : MyEnum.values()[0].forValue(value)));
```
andreaTP, line 139: I think this is a fair assumption/implementation to get an instance from the enum.

baywet: why not simply MyEnum::forValue with a static method instead?

andreaTP: Because it cannot be tracked by the type system, to use `MyEnum::forValue` we need `forValue` to be a static method.

baywet: which is used to be until the recent change?

andreaTP: Correct, but in this way we can call the `forValue` without going through reflection which is the point of this PR.
The alternative to this lambda would be to generate another static method to return the enum parser, the change will happen in the generator codebase.

baywet: Sorry, I still don't understand why we can't simply pass the reference to the static method, without having to use reflection? Like we do for error mappings and so

andreaTP: Thanks for insisting, I was completely overlooking it, changing things accordingly now

baywet (approved):

Issue #923, opened 2023-12-06
Remove usage of reflection from `abstractions`


andreaTP:
The libraries are currently using reflection in 4 different places (if I'm not missing something):

- `ApiClientBuilder` - `abstractions`
- `RequestInformation` - `abstractions`
- `KiotaSerialization` - `abstractions`
- `JsonSerializationWriter` - `serialization/json`

As ppl can roll their json serialization support I'm not concerned with the latter.
But, the usage of reflection in `abstractions` means that this is not going to run without configuration when compiled with [`native-image`](https://www.graalvm.org/22.0/reference-manual/native-image/), especially for the usage in Quarkus applications.

Since we are in control of the generator, I do believe that we can find solutions to completely avoid the usage of reflection in the `abstractions` layer.

baywet:
Thanks for starting this.
Yes this should be easy to remove from the builder, and you started the work.
For request information, this should only be present for the query parameters, and maybe for the request options, and I'm not sure there is a way around that?
For serialization, the methods have overloads where you can pass the factory method to avoid using reflection already, so we should be good here.

andreaTP:
> For request information, this should only be present for the query parameters, and maybe for the request options, and I'm not sure there is a way around that?

We should be able to generate(as source) a static method to retrieve the field list and the needed information with a plain method call. Would you consider such an approach?

> For serialization, the methods have overloads where you can pass the factory method to avoid using reflection already, so we should be good here.

Agreed, I'll review this detail before closing the issue.

baywet:
> We should be able to generate(as source) a static method to retrieve the field list and the needed information with a plain method call. Would you consider such an approach?

I'd like to explore that in more details:
If the method just returned a list of field names, we'd still need reflection to get the values + serialization name.
If the method returned directly the query parameters key + value, it'd still need reflection to get to the attribute in case the query parameter serialization name is different from the symbol name.
I don't see benefits from pushing that to the code gen, and it comes at a size price.
In dotnet, native compilation relies on attributes that basically mean "yes you can trim things, except for the argument to this method, because the method will list out the fields of the passed type" giving us the best of both worlds. Is there something similar to that in Java native?

andreaTP:
> Is there something similar to that in Java native?

Kind of, but pretty tricky to setup: https://www.graalvm.org/jdk17/reference-manual/native-image/dynamic-features/Reflection/#manual-configuration and this will need to be done for the generated code ...

[...]

PR #957, opened 2024-01-03
[chore] Remove reflection usage from KiotaSerialization

andreaTP:
I started picking up #923 again.
Before opening Java to GA I think it makes sense to remove those reflective methods, they look convenient, but I think that a cleaner smaller API is the way to go to start with.

Also, this will prevent accidental usage of reflection on platforms that do not support it completely (like GraalVM native-image).

In case we receive feedback and ask to re-introduce those helpers we can re-evaluate.


baywet (changes requested):
Thanks for the contribution.
The reflection option was requested by @sebastienlevert in the specification so people don't have to provide the factory method. It's a shorthand for simplicity and since there are overloads accepting the method I don't think it hurts to have those.
They are not used by the generated code.

baywet:
These are most of the arguments I gave to @sebastienlevert + reliability/performance. I'll let you folks discuss that since I'm already in the camp of "we should not be doing any reflection"

andreaTP:
Got it 👍 
Personally, the only "unusual" detail in the reflection-free API is the name of the method IMHO: `createFromDiscriminatorValue` which is rather alien terminology for Java developers.

The resulting API looks cleaner with a more "idiomatic" name like:
```java
KiotaSerialization.deserialize(jsonContentType, strValue, TestEntity::deserializer);
```

sebastienlevert:
In this case @andreaTP if we were to remove the reflection versions of these APIs, would the only thing left on your mind be the `deserializer` vs. `createFromDiscriminatorValue`? That even makes me think that this "wording" might be better in other languages but would impact our GA languages. @baywet any opinions on renaming this factory?

andreaTP:
> if we were to remove the reflection versions of these APIs

I would say that no further actions are required around reflection, all the rest falls in the category of "nice to have".

sebastienlevert:
So current state is OK? Just to make sure I understand correctly?

andreaTP:
Correct, the current status is OK and Kiota is compatible with `native-image` already.

To avoid introducing breaking changes after GA, I warmly encourage you to approve this PR, and we can move on from there.

andreaTP:
Any update? Should I rebase or this is not going to go in?

sebastienlevert:
This should go in. I agree with your concerns and better to avoid reflection in this case.

@baywet make your magic happen 😁

baywet:
Just a heads up on that @sebastienlevert : this means that Java will be different from dotnet. But similar to Go/TS. But I'll proceed. Thanks everyone for the discussion.

baywet (approved):

PR #991, opened 2024-01-09
Reflection free query parameters expansion

andreaTP:
Fix for: https://github.com/microsoft/kiota/issues/3965


components/abstractions/src/test/java/com/microsoft/kiota/RequestInformationTest.java
```diff
@@ -310,18 +312,23 @@ void ReplacesEnumValuesPathParameters() throws IllegalStateException, URISyntaxE
...
     @jakarta.annotation.Nullable public String search;
 
-    @QueryParameter(name = "dataset")
     @jakarta.annotation.Nullable public TestEnum dataset;
 
-    @QueryParameter(name = "datasets")
     @jakarta.annotation.Nullable public TestEnum[] datasets;
+
+    @jakarta.annotation.Nonnull public Map<String, Object> toQueryParameters() {
+        Map<String, Object> allQueryParams = new HashMap();
```
baywet, line 327:
```suggestion
        final Map<String, Object> allQueryParams = new HashMap();
```

components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
```diff
@@ -152,29 +151,19 @@ public void setUri(@Nonnull final URI uri) {
...
+        Map<String, Object> params = parameters.toQueryParameters();
+        for (Map.Entry<String, Object> entry : params.entrySet()) {
+            if (entry.getKey() != null && entry.getValue() != null) {
+                Object value = entry.getValue();
                 if (value != null) {
                     value = replaceEnumValue(value);
                     if (value.getClass().isArray()) {
-                        queryParameters.put(name, Arrays.asList((Object[]) value));
+                        queryParameters.put(entry.getKey(), Arrays.asList((Object[]) value));
                     } else if (!value.toString().isEmpty()) {
```
baywet, line 163: we might want to switch to an else (remove the condition) because of this discovery

https://github.com/microsoft/kiota-abstractions-dotnet/issues/172

baywet (changes requested):
Thanks for taking this on. A few minor comments.

components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
andreaTP: If I understand it correctly this is your ask:
https://github.com/microsoft/kiota-java/pull/991/commits/75a24aa3067c8a20dcbd18fa74f72cf70650f367

baywet: yep, this is another remnant of the early days of kiota. For some reason we didn't test for `?queryParam=` or even just `?queryParam` which are both valid scenarios. Thanks for taking this on. Resolving.

baywet (approved):

components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
andreaTP: imagined, is change small enough that makes sense to include since I'm touching those methods 👍

components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
```diff
@@ -152,29 +151,19 @@ public void setUri(@Nonnull final URI uri) {
...
-            try {
-                Object value = field.get(parameters);
-                String name = field.getName();
-                if (field.isAnnotationPresent(QueryParameter.class)) {
-                    final String annotationName = field.getAnnotation(QueryParameter.class).name();
-                    if (annotationName != null && !annotationName.isEmpty()) {
-                        name = annotationName;
-                    }
-                }
+    public void addQueryParameters(@Nullable final QueryParameters parameters) {
```
andreaTP, line 154: argh, forgot a line here:

```
if (parameters == null) return;
```

baywet: arg, I missed it too... Please submit a patch (with patch bump for the version) as I just finished releasing it.

andreaTP: noticed, and it's ready 😄

PR #992, opened 2024-01-09
Consider the case where QueryParameters is null

andreaTP:
@baywet forgot this line 😅


baywet (approved):
Thanks for noticing and for the quick patch

andreaTP:
Sorry for the noise ...

Issue #939, opened 2023-12-14
Using https:// as part of the allowedHosts shouldn't be an issue


maisarissi:
When using Azure Identity to create an authProvider and adding a allowedHost as the following:

```java
final String[] allowedHosts = new String[] {"https://graph.microsoft.com"};
```
makes the client creation fail with "empty token" error.

I should be able to add allowedHosts with and without https://

andrueastman:
@baywet  Should we also move this one up to include all the other languages' `AllowedHostValidators` to try to strip out URI semantics before comparing the hosts?

baywet:
@andrueastman since this is not impacting generation I meant to create issues in the other abstractions repos. But I had forgotten. This is now done, thanks for the nudge. This should be considered a low priority item for now, and http (without the S) should also be considered, as well as different casings.

andrueastman:
> @andrueastman since this is not impacting generation I meant to create issues in the other abstractions repos. But I had forgotten. This is now done, thanks for the nudge. This should be considered a low priority item for now, and http (without the S) should also be considered, as well as different casings.

Thanks for confirming here

XdpCs:
I believe there may not be a need to make changes to this issue, as it introduces potential confusion regarding the concept of hosts.

baywet:
@XdpCs can you expand on the scenario you are thinking about please?

XdpCs:
In my opinion, allowedHosts should only include the hosts, not the entire URLs. Including ‘http’ or ‘https’ would involve the entire URL, while hosts are just a part of the URL.

baywet:
I understand the concern about the naming. The reason why Maisa created this issue is because we've observed it's a mistake (leave the scheme) that's commonly made and often trips people when they first try kiota.
Instead of stripping it automatically, we could throw an exception with an explicit message. What do you think about this alternative?

XdpCs:
I appreciate the consideration of this alternative. However, if someone directly utilizes [the specified link](https://github.com/microsoft/kiota-abstractions-go/blob/6f6c124cea15fa90d1a96763fbfbc01259710bfa/authentication/allowed_hosts_validator.go#L14) in the kiota-abstractions-go repository, there might be compatibility issues. This function will add an error param for return.

XdpCs:
If you don't think this is a problem, I will provide the Go version.

PR #1051, opened 2024-01-31
Validate allowed hosts for http and https prefixes

Ndiritu:
closes https://github.com/microsoft/kiota-java/issues/939

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Removed methods using reflection from `KiotaSerialization`
+- Improve `AllowedHostsValidator` to throw an error if `https://` or `http://` prefix is present in a allowed host value.
 
 ## [0.12.1] - 2024-01-10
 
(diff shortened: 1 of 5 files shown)


components/abstractions/src/main/java/com/microsoft/kiota/authentication/AllowedHostsValidator.java
```diff
@@ -40,8 +40,13 @@ public void setAllowedHosts(@Nonnull final Set<String> allowedHosts) {
         validHosts = new HashSet<String>();
         if (allowedHosts != null) {
             for (final String host : allowedHosts) {
-                if (host != null && !host.isEmpty())
+                if (host != null && !host.isEmpty()) {
+                    if (host.startsWith("http://") || host.startsWith("https://")) {
```
baywet, line 44: We need the normalization (lower case, trim) to happen before we check for the presence of a scheme

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Removed methods using reflection from `KiotaSerialization`
+- Improve `AllowedHostsValidator` to throw an error if `https://` or `http://` prefix is present in a allowed host value.
```
baywet, line 14: You can add the version entry directly

baywet (changes requested):
Thanks for taking this on. A couple of remarks

baywet (approved):

PR #1070, opened 2024-02-06
WIP: Untyped nodes 

andrueastman:
Related to https://github.com/microsoft/kiota/issues/2319

### DIFF
--- a/components/abstractions/src/main/java/com/microsoft/kiota/serialization/UntypedNode.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/serialization/UntypedNode.java
@@ -1,0 +1,44 @@
+package com.microsoft.kiota.serialization;
+
+import jakarta.annotation.Nonnull;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Consumer;
+
+/**
+ * Base class for untyped node.
+ */
+public class UntypedNode implements Parsable {
+
+    /**
+     * The deserialization information for the current model.
+     * @return The map of serializer methods for this object.
+     */
+    @Nonnull @Override
+    public Map<String, Consumer<ParseNode>> getFieldDeserializers() {
+        return new HashMap<>();
+    }
+
+    /**
+     * Serializes the current object
+     */
+    @Override
+    public void serialize(@Nonnull SerializationWriter writer) {}
+
+    /**
+     * Gets the value assigned to untyped node.
+     * @return The value assigned to untyped node.
+     */
+    public Object getValue() {
+        return null;
+    }
+
+    /**
+     * Creates a new instance of the appropriate class based on discriminator value.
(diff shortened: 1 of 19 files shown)


components/abstractions/src/main/java/com/microsoft/kiota/serialization/UntypedNode.java
```diff
@@ -0,0 +1,48 @@
...
+    public void serialize(@Nonnull SerializationWriter writer) {
+        // no properties to serialize. This is handled by custom serialization logic.
+    }
+
+    /**
+     * Gets the value assigned to untyped node.
+     * @return The value assigned to untyped node.
+     */
+    @Nullable public Object getValue() {
+        return null;
```
baywet, line 37: should this throw a not implemented exception or equivalent to align the implementation with dotnet?

components/serialization/json/src/main/java/com/microsoft/kiota/serialization/JsonParseNode.java
```diff
@@ -208,10 +209,46 @@ private <T> List<T> iterateOnArray(Function<JsonParseNode, T> fn) {
...
+        return getUntypedValue(currentNode);
+    }
+
+    @Nonnull private UntypedNode getUntypedValue(JsonElement element) {
+        if (element.isJsonNull()) return new UntypedNull();
+        else if (element.isJsonPrimitive()) {
+            final JsonPrimitive primitive = element.getAsJsonPrimitive();
+            if (primitive.isBoolean()) return new UntypedBoolean(primitive.getAsBoolean());
+            else if (primitive.isString()) return new UntypedString(primitive.getAsString());
+            else if (primitive.isNumber()) return new UntypedDouble(primitive.getAsDouble());
```
baywet, line 227: when are we effectively using demical/float/long/integer? I know it can be hard to tell without the requested type and solely from from the data.

andrueastman: I believe I did this because of the challenge you speak of. A similar implementation observed in the `tryGetAnything()` method at https://github.com/microsoft/kiota-java/blob/41bafdbc61d0ee8cc36de993874e4bd793024dda/components/serialization/json/src/main/java/com/microsoft/kiota/serialization/JsonParseNode.java#L276
I ended up keeping the other types as they can still be used during serialization though.

baywet: The only thing I'm a bit worried about is the symmetrical aspect of it.
If somebody serializes as int, they'd probably expect to get an int back.
But I guess as long as this choice is consistent across languages, we can re-evaluate later based on customer feedback.

baywet (approved):

PR #1084, opened 2024-02-13
Add default UTC offset when deserializing to OffsetDateTime fails due to a missing time offset value.

Ndiritu:
Checks for failure case by parsing to [LocalDateTime](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html) (doesn't contain time offset), otherwise returns initial OffsetDateTime parsing exception.

closes https://github.com/microsoftgraph/msgraph-sdk-java/issues/1785


baywet (approved):
Thanks for fixing this!

PR #1088, opened 2024-02-17
Content Length Missing Fix 

ramsessanchez:
There have been several issues raised concering the failure of FileUpload. The issue stems from the fact that we did not include an `Override` on the contentLength method on the RequestBody custom constructor when translating a RequestInformation object to an Okhttp Request object. I have also added a property on RequestInformation that is set whenever the `content` property is set on the RequestInformation object.  While we did include a `content-length` header it seems to also be a requirement that this property is set on the RequestBody otherwise we will receive an error stating that the 'content length is unknown'. 

This will lead to version 1.0.3


andrueastman (approved):

components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
```diff
@@ -268,26 +271,31 @@ public void setResponseHandler(@Nonnull ResponseHandler responseHandler) {
     /**
      * Sets the request body to be a binary stream.
      * @param value the binary stream
+     * @throws IOException when the stream cannot be read.
      * @deprecated use {@link #setStreamContent(InputStream, String)} instead.
      */
     @Deprecated
-    public void setStreamContent(@Nonnull final InputStream value) {
+    public void setStreamContent(@Nonnull final InputStream value) throws IOException {
```
baywet, line 278: Adding new checked exceptions will be source breaking for the generated code.
(Requiring the exception to be handled, or be declared at a higher level)
We should probably wrap this to maintain compatibility.
@andreaTP to provide additional input if he can

baywet (changes requested):

components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
```diff
@@ -268,26 +271,31 @@ public void setResponseHandler(@Nonnull ResponseHandler responseHandler) {
...
-            @Nonnull final InputStream value, @Nonnull final String contentType) {
+            @Nonnull final InputStream value, @Nonnull final String contentType)
+            throws IOException {
         Objects.requireNonNull(value);
         Objects.requireNonNull(contentType);
         if (contentType.isEmpty()) {
             throw new IllegalArgumentException("contentType cannot be empty");
         }
         this.content = value;
+        this.contentLength = (long) value.available();
```
andreaTP, line 297: I'm not sure this is correct, since this is public API and we are not in control of the argument `available` might have very different meanings:

https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#available--

andreaTP (changes requested):
I'm not concerned about breaking the API in this case, as almost anything we are going to do on the `InputStream` is gonna throw an `IOException`.

I believe that relying on `available` is **not** the correct resolution for this bug.

Having a test that stresses the use case would be great.

components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
calebkiage: I read in _Effective Java_ that checked exceptions should be used for recoverable errors and runtime exceptions for programming errors. I think failing to set the stream content may not be recoverable for that specific request.

components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
```diff
@@ -268,26 +271,31 @@ public void setResponseHandler(@Nonnull ResponseHandler responseHandler) {
...
 
     /**
      * Sets the request body to be a binary stream.
      * @param value the binary stream
      * @param contentType the content type of the stream.
+     * @throws IOException when the stream cannot be read.
      */
     public void setStreamContent(
-            @Nonnull final InputStream value, @Nonnull final String contentType) {
+            @Nonnull final InputStream value, @Nonnull final String contentType)
```
calebkiage, line 289: I think this method should either accept a content length or have an overload that accepts a content length.

```java
public void setStreamContent(@Nonnull final InputStream value, @Nonnull final String contentType, final Long contentLength) {...}
```

components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
baywet: yes, ideally we would have set this checked exception in the method prototype before the GA.
but since we GAed the language, adding this now would be a source breaking change.
i.e.: any client that's been generated to date will break if they pull the latest kiota dependencies without re-generating the client.

components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
baywet: instead of making changes to request information that are specific to java, and have undesired effects (source breaking or wrapping exceptions), could we handle that in the request adapter? (see my other comment)

calebkiage: Adding an overload isn't a source breaking change unless I'm mistaken. If it works in the request adapter as well, I'm fine with it. But I'd argue that content information belongs in the request information since it's part of the request.

baywet: right, but it'd be the only language with a content length field. I'm not sure why we'd do that only for Java?

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
```diff
@@ -882,6 +882,16 @@ private void setBaseUrlForRequestInformation(@Nonnull final RequestInformation r
                     requestInfo.content == null
                             ? null
                             : new RequestBody() {
+                                @Override
+                                public long contentLength() {
+                                    if (requestInfo.contentLength != null) {
+                                        spanForAttributes.setAttribute(
+                                                "http.request_content_length",
+                                                requestInfo.contentLength);
+                                        return requestInfo.contentLength;
```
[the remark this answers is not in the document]
ramsessanchez: I initially thought that it would make most sense to add it to the requestInformation via the suggestion to put in an overload, but seeing as it is a very commonly exposed method I agree that adding the check at the request adapter level and handling the exception there would be the least breaking method.

components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
ramsessanchez: While I do agree that adding the check at the RequestAdapter makes sense as the least interruptive method for handling this, I also agree with @calebkiage when we consider that we already include contentLength in a lot of the methods for java specifically due to the languages need for this value across many operations when dealing with input streams in particular. I'm not familiar with the other languages and their handling of this, but with java it seems the content-length appears frequently and thus is a valid candidate for a value that should be stored in the requestInformation specifically for the Java language.

ramsessanchez:
I have refactored the code to utilize Okhttp's method RequestBody.create(Byte[], MediaType)
Using the functionality made available in Compatability.readAllBytes[] we can convert the inputStream to a byte array which we then pass to the create function, this removes the need to Overwrite any of the methods and rather use OkHttp's native implementation. Further, by converting the input stream to a byte[] array we should be able to more reliably get the length of the stream without using the `available` method which @andreaTP pointed out can have unwanted affects if the inputStream does not correctly implement the method. I have wrapped the ioException in a runtime exception, though that is temporary as I would like to hear everyone's thoughts as to how best we should handle this exception.

araneolus:
With this solution, is it no longer necessary to know the length of the InputStream in advance? The background is that in my program I first write the InputStream into a ByteArray to determine the final length, which I do not know, because I get the data from different data sources.
So far I have to set the length in the "DriveItemUploadableProperties". 
If now the OkHttp method does the same, I have a doubling of the data in the memory which is not exactly resource-saving.

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
```diff
@@ -878,32 +875,24 @@ private void setBaseUrlForRequestInformation(@Nonnull final RequestInformation r
...
+                                : new HashSet<>();
+                final MediaType mediaType;
+                if (!contentTypes.isEmpty()) {
+                    String contentType = contentTypes.toArray(new String[] {})[0];
+                    spanForAttributes.setAttribute("http.request_content_type", contentType);
+                    mediaType = MediaType.parse(contentType);
+                } else {
+                    mediaType = null;
+                }
+                byte[] bytes = Compatibility.readAllBytes(requestInfo.content);
```
calebkiage, line 892: Isn't this a potential problem for large files? Imagine a 30GB upload to a cloud drive. This would require a 30GB allocation, no?

calebkiage: The reason I think could be a problem is because, for computers with lower RAM than the content size, the program will either crash (Allocation failure) or swap into disk if swapping is enabled (which will be a lot slower)

baywet: yes this is going to cause memory pressure, which is going to cause the application to be at best slow at worst unstable. we should review our approach here.

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
```diff
@@ -878,32 +875,24 @@ private void setBaseUrlForRequestInformation(@Nonnull final RequestInformation r
...
-                                        return MediaType.parse(contentType);
-                                    }
-                                }
-
-                                @Override
-                                public void writeTo(@Nonnull BufferedSink sink) throws IOException {
-                                    sink.writeAll(Okio.source(requestInfo.content));
-                                }
-                            };
+            RequestBody body = null;
```
baywet, line 878: I think we should revert how we build the request body.
HTTP-wise there are two cases for the request body and the content length header:
- the body's length is known in advance, and should be set in the header, so the server knows "when to stop reading". This is typically the case of a JSON payload that comes from serializing models etc...
- the body's length is not known in advance, in which case the header should not be set. The client will be streaming the bytes and using special semantics in the body to indicate it's done streaming the bytes. The is typically the case of a large file upload (not referring to the Microsoft Graph Task), and this is done this way because we CAN'T know in advance the size, and don't want to load everything in memory to avoid memory pressure.

We do have two requirements that come from this:
- setting the telemetry attribute for the content length: We can do this by [reading the content length field of the request body](https://square.github.io/okhttp/3.x/okhttp/okhttp3/RequestBody.html#contentLength--) and if it's not -1, set it, otherwise, ignore it. The RequestBody WriteTo method will ensure the content length is provided when available, exactly like [the dotnet implementation](https://github.com/microsoft/kiota-http-dotnet/blob/30d1d1f70e06498735feba01e7aba71b4271409b/src/HttpClientRequestAdapter.cs#L569)
- providing the input stream to the request body, so it can be streamed, and so it can provide the content length information both in the request it's sending, and to the prior point. This was already implemented with the writeTo method implementation as far as I understand.

Now, it it possible the content length information is never available, but I'd rule it a bad implementation from ok http, and even if that is the case, all of our serialization writers actually return https://docs.oracle.com/javase/8/docs/api/java/io/ByteArrayInputStream.html under the covers, We could downcast the input stream to then read the available method's return, which should give us the right value.

I hope this clarifies things for everyone?

baywet: additionally, we might also want to read the Content-Type request information header (from the headers collection) when present.

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
ramsessanchez: For the case of our SDK we have a limit of 5mb chunks, perhaps we can set this as the upper limit here as well?

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
ramsessanchez: @baywet I am having trouble understanding if the contentLength() method is called before or after the writeTo() method, the documentation isn't super clear and it sort of reads as if it is done prior to writeTo(). I think downcasting could work, but that should perhaps be a back up plan. As you mentioned, given our implementation for largeFileUpload we do include the contentLength header, so we should check that value if present, return it if it is, and downcast and check available if that value is not included. Would that make sense?

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
baywet: no we shouldn't, nothing prevents somebody from generating a client to an API that accepts a stream. And read that stream from the local file system (or from another network call). Limiting things here to 5MB would effectively block any scenario about 5MB.

For the SDK, and for others reading, we're referring to 5MB *slices* (chunk encoding is something different in HTTP) for the large file upload task, which is specific to Microsoft Graph.

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
baywet: after the writeTo because the value will only be populated once the stream has been enumerated as far as I can tell.
yes I guess the rationale should be:
1. try to read the content-length header from the request information, if present override the implementation of the content length method in request body (if possible)
2. read the content length method only once the request body object has been built, if we get any other value than -1, set it for telemetry.

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
```diff
@@ -899,6 +900,31 @@ public MediaType contentType() {
...
+                                    final Set<String> contentLength =
+                                            requestInfo.headers.getOrDefault(
+                                                    contentLengthHeaderKey, new HashSet<>());
+                                    if (contentLength.isEmpty()) {
+                                        try (final ByteArrayInputStream contentStream =
+                                                (ByteArrayInputStream) requestInfo.content) {
+                                            length = contentStream.available();
+                                        } catch (IOException ex) {
+                                            length = -1L;
+                                        }
```
baywet, line 915:
```suggestion
                                    if (contentLength.isEmpty() && requestInfo.content instanceof ByteArrayInputStream) {
                                        final ByteArrayInputStream contentStream = (ByteArrayInputStream) requestInfo.content;
                                        try  {
                                            length = contentStream.available();
                                        } catch (IOException ex) {
                                            length = -1L;
                                        }
```

we don't want to be disposing the stream here since it's already taken care of by okhttp down the pipeline
we also want to type test before we cast to avoid having things blow up if it's not the expected type.

baywet (changes requested):
Get close! I'd like unit tests to be implemented for both scenarios:
- the stream is coming from local file system
- the stream is a known JSON payload

To make sure the headers are what we expect

baywet (dismissed):

baywet (dismissed):

baywet (approved):

araneolus:
Hooray, it works :-)

araneolus:
Unfortunately I have to take back my Hooray :-(
The following error occurs with large files.

_Can't write 'testfolder/jdk1.6.0_45.tgz' Range [5242880, 5242880 + 5242880) out of bounds for length 5242880_

I am setting the size of the File at driveItemUploadableProperties

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
```diff
@@ -899,6 +900,30 @@ public MediaType contentType() {
...
+                                    long length;
+                                    final Set<String> contentLength =
+                                            requestInfo.headers.getOrDefault(
+                                                    contentLengthHeaderKey, new HashSet<>());
+                                    if (contentLength.isEmpty()
+                                            && requestInfo.content
+                                                    instanceof ByteArrayInputStream) {
+                                        final ByteArrayInputStream contentStream =
+                                                (ByteArrayInputStream) requestInfo.content;
+                                        length = contentStream.available();
```
andreaTP, line 914: I'm so sorry I have not followed up on this subject, but, as noticed here: https://github.com/microsoft/kiota-java/pull/1088#discussion_r1494670806

`available` is *NOT* the API you need to use in this context since it has very very different semantics from the expected behavior.
Have you explored different solutions?
Do we have a reproducer of the original issue that triggered those changes?

andreaTP: Unfortunately, due to the support and semantics, I'm using workarounds in the Apicurio Registry testsuite to properly perform file uploads.
But I'm looking forward to a real resolution of the issue.

If you need me to chip in I would love a minimal repoducer, maybe @araneolus would be able to provide one 🙏

components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
andreaTP: ☝️

araneolus:
Ok, This is my Test code.
I have two issues with this code.

- If I use the tempFilePath and the Data is smaller 4M, the method "sendPrimitive" is running on a Exception under windows "_the process cannot access the file because it is being used by another process_" It works well with the ByteArray tempBuffer.
- If the file is larger than 4M I got the error "_Can't write 'testfolder/jdk1.6.0_45.tgz' Range [5242880, 5242880 + 5242880) out of bounds for length 5242880_"


```
public void put(final OFSFile<?> ofsFile, final String destination, final InputStream fis) throws IOException {
	try {
		connect();
	} catch (InterruptedException e) {
		throw new IOException("Interrupted");
	}
	Drive drive = this.getDrive();
	if (drive == null) {
		throw new IOException("Drive is empty");
	}
	Path tempFilePath = null;
	long streamSize = 0;
	byte[] tempBuffer = null;
	if (this.isTempFileUse()) {
		tempFilePath = Files.createTempFile("vec", ".tmp");
		try {
			Files.copy(fis, tempFilePath);
			streamSize = Files.size(tempFilePath);
		} catch (Exception e) {
			log.error("[put] " + this.getSessionId() + " Can't create tempfile '" + tempFilePath.getFileName()
					+ "' " + e.getMessage());
			throw new IOException("Can't create tempfile '" + tempFilePath.getFileName() + "' " + e.getMessage());
		} finally {
			fis.close();
		}
		streamSize = Files.size(tempFilePath);
	} else {
		tempBuffer = IOUtils.toByteArray(fis);
		streamSize = tempBuffer.length;
	}
	InputStream inputstream = null;
	try {
		if (this.isTempFileUse()) {
			inputstream = Files.newInputStream(tempFilePath, java.nio.file.StandardOpenOption.READ);
		} else {
			inputstream = new ByteArrayInputStream(tempBuffer);
		}
		IProgressCallback progressCallback = new IProgressCallback() {
			@Override
			public void report(final long current, final long max) {
				log.debug(String.format("Uploaded %d bytes of %d total bytes", current, max));
			}
		};
		if (streamSize < 4194304) {
			RequestInformation requestInformation = graphClient.drives().byDriveId(drive.getId()).items()
					.byDriveItemId("root:/" + destination + ":").content().toPutRequestInformation(inputstream);
			URI uriIncludesConflictBehavior = new URI(
					requestInformation.getUri().toString() + "?@microsoft.graph.conflictBehavior=replace");
			requestInformation.setUri(uriIncludesConflictBehavior);
			graphClient.getRequestAdapter().sendPrimitive(requestInformation, null, InputStream.class);
		} else {

			DriveItemUploadableProperties driveItemUploadableProperties = new DriveItemUploadableProperties();
			driveItemUploadableProperties.setName(FilenameUtils.getName(destination));
			driveItemUploadableProperties.setFileSize(streamSize);

			Map<String, Object> additionalData = new HashMap<>();
			additionalData.put("@microsoft.graph.conflictBehavior", "replace");
			driveItemUploadableProperties.setAdditionalData(additionalData);

			Map<String, Object> commitData = new HashMap<>();
			commitData.put("deferCommit", false);
			
			CreateUploadSessionPostRequestBody uploadSessionPostRequestBody = new CreateUploadSessionPostRequestBody();
			//uploadSessionPostRequestBody.setAdditionalData(commitData);
			uploadSessionPostRequestBody.setItem(driveItemUploadableProperties);
			
			UploadSession uploadSession = null;
			try {
				RequestInformation test = graphClient.drives().byDriveId(drive.getId()).items()
						.byDriveItemId("root:/" + destination + ":").createUploadSession()
						.toPostRequestInformation(uploadSessionPostRequestBody);
				String xx = IOUtils.toString(test.content, StandardCharsets.UTF_8);
				
				uploadSession = graphClient.drives().byDriveId(drive.getId()).items()
						.byDriveItemId("root:/" + destination + ":").createUploadSession()
						.post(uploadSessionPostRequestBody);
			} catch (ApiException gse) {
				if (gse.getResponseStatusCode() == 404) {

				}
			}

			LargeFileUploadTask<DriveItemUploadableProperties> uploadTask = new LargeFileUploadTask<>(
					graphClient.getRequestAdapter(), uploadSession, inputstream, streamSize,
					DriveItemUploadableProperties::createFromDiscriminatorValue);

			try {
				UploadResult<DriveItemUploadableProperties> uploadResult = uploadTask
						.upload(this.getRetry() <= 0 ? 5 : this.getRetry(),progressCallback);
				if (uploadResult.isUploadSuccessful()) {
					log.info("[put] " + this.getSessionId() + "finished. File '" + destination
							+ "' succcessfully written");
				}
			} catch (ApiException | InterruptedException exception) {
				log.error("[put] " + this.getSessionId() + "Exception :" + exception.getMessage());
				exception.printStackTrace();
			}
		}
	} catch (Exception e) {
		log.error("[put] " + this.getSessionId() + " Can't write '" + destination + "' " + e.getMessage());
		throw new IOException("Can't write '" + destination + "' " + e.getMessage());
	} finally {
		if (tempFilePath != null) {
			Files.delete(tempFilePath);
		}
		if (inputstream != null) {
			inputstream.close();
		}
		tempBuffer = null;
	}
	return;
}

```

Issue #1110, opened 2024-03-05
solving for content-length request header


baywet:
Hey @Ndiritu @calebkiage @andreaTP @ramsessanchez @araneolus @ffcdf  (I'm probably missing others, feel free to mention them as well)
We're having conversations about the content-length all over the place, and I think this is not helping solving the issues for good.

Here are the different contexts where the content-length is critical from an HTTP standpoint:
- When the length is known in advance (small payloads, large file upload implementation), it MUST be set.
- When the length is NOT known in advance (streaming) it MUST NOT be set.

From there, we have a couple of additional requirements/constraints:
- OkHttp will set the content-length header automatically, once the writeTo method is called on the body, depending on the media type and the stream type from what I can tell. But the rules are unclear.
- Request information instances need to be able to override the Content-Length header (large file upload implementation mostly)
- The content length value needs to be passed to the span attribute for tracing when it's available.

@andreaTP rightfully pointed out that [we should not be using the available method](https://stackoverflow.com/a/1119346/3808675) since implementations behaviour vary. I think that part of the confusion here is that streams in dotnet behave a bit differently w.r.t. that aspect.
https://github.com/microsoft/kiota-java/blob/41bafdbc61d0ee8cc36de993874e4bd793024dda/components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java#L914

So one first step we probably need to take is to remove that if block body. (and yes, I know I'm the one [who made that suggestion](https://github.com/microsoft/kiota-java/pull/1088#discussion_r1499161328), maybe I should've taken more sick days)

I already fixed a regression to restore the default okhttp behaviour when the content length header is not overridden by the request in #1109 (maybe the code here needs to be updated to be a strict rather than lower or equal comparison, and set the initial value to -1)

This block most likely needs to be moved back to after the body is built because we won't get the native writeTo behaviour otherwise.
https://github.com/microsoft/kiota-java/blob/41bafdbc61d0ee8cc36de993874e4bd793024dda/components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java#L923
(but it needs to be tested thoroughly)

[...]

PR #1119, opened 2024-03-18
Fix/content length

Ndiritu:
Only override content-length with values from Content-Length header or if the body is a `ByteArrayInputStream` since the [`available()`](https://docs.oracle.com/javase/8/docs/api/java/io/ByteArrayInputStream.html#available--) method is reliable from the array size backing the stream.

closes https://github.com/microsoft/kiota-java/issues/1110
closes https://github.com/microsoftgraph/msgraph-sdk-java/issues/1883

Pending:
- [x] Integration testing
- [x] Large File Upload Scenarios

### DIFF
--- a/components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
+++ b/components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
@@ -907,30 +907,20 @@ public class OkHttpRequestAdapter implements com.microsoft.kiota.RequestAdapter 
 
                                 @Override
                                 public long contentLength() throws IOException {
-                                    long length;
                                     final Set<String> contentLength =
                                             requestInfo.headers.getOrDefault(
                                                     contentLengthHeaderKey, new HashSet<>());
-                                    if (contentLength.isEmpty()
-                                            && requestInfo.content
+                                    if (!contentLength.isEmpty()) {
+                                        return Long.parseLong(
+                                                        contentLength.toArray(new String[] {})[0]);
+                                    }
+                                    if (requestInfo.content
                                                     instanceof ByteArrayInputStream) {
                                         final ByteArrayInputStream contentStream =
                                                 (ByteArrayInputStream) requestInfo.content;
-                                        length = contentStream.available();
-                                    } else {
-                                        length =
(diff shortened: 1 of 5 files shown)


components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
```diff
@@ -907,30 +906,19 @@ public MediaType contentType() {
...
-                                    }
-                                    if (length <= 0) {
-                                        length = super.contentLength();
-                                    }
-                                    if (length > 0) {
-                                        spanForAttributes.setAttribute(
-                                                HttpIncubatingAttributes.HTTP_REQUEST_BODY_SIZE,
-                                                length);
+                                        return contentStream.available();
                                     }
```
baywet, line 920: From the "summarized" issue
> @andreaTP rightfully pointed out that [we should not be using the available method](https://stackoverflow.com/a/1119346/3808675) since implementations behaviour vary. I think that part of the confusion here is that streams in dotnet behave a bit differently w.r.t. that aspect.

I haven't tested the latest version of the code at this point. And it's likely that in most cases a byte array input stream which hasn't been read will return the total payload length when reading available. But to be safe, are we sure we absolutely need this block since super.contentLength has been restored?

Ndiritu: Tests fail because `super.contentLength()` sets [`-1`](https://github.com/square/okhttp/blob/master/okhttp/src/main/kotlin/okhttp3/internal/-RequestBodyCommon.kt#L43) by default on the Request object. 

Okhttp probably always adds the right content-length when no overrides are used but I'm unable to unit test for it, even at the interceptors, the Content-Length is not yet added as a header. Might you know how else I can test?

baywet: thanks for testing this. Can you at least add a comment for context for this block please?
And no, I'm not aware of any other way of testing besides maybe setting up a mock server in the unit test and testing what the mock server receives. But even then, the telemetry attribute would always be wrong...

baywet (approved):

PR #1130, opened 2024-03-19
Fix UnsupportedOperationException when not providing scopes to AzureIdentityAccessTokenProvider

joakibj:
Fixes microsoftgraph/msgraph-sdk-java#1882

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


joakibj:

andrueastman (approved):

baywet:
update: I've done the changelog changes, only the other points are left

joakibj:
Please see my latest update
- [x] update l132 to 135 to work on a new array instead of the field, so the implementation is identical to the dotnet one
- [x] update the version accordingly in gradle.properties and in the user agent options

There is also a difference from the dotnet implementation at line 66, as it seems [dotnet does case insensitive ordering of the scopes](https://github.com/microsoft/kiota-authentication-azure-dotnet/blob/c93b68028d607dcccfc58e9e6750a65b26be2e09/src/AzureIdentityAccessTokenProvider.cs#L46). But I believe this is out of scope 😎 for this pull request / issue

baywet (approved):

PR #1149, opened 2024-04-02
Remove unnecessary @Nullable annotations

### LINKED ISSUE #1145: Request headers shouldn't be `@Nullable`
First of all, thanks for your huge work on this project :heart: 

I have a question about `@Nullable` in the `headers` field in the `BaseRequestConfiguration` class.

In what situations could this field be nullable? This requires API consumers to implement null checks, as it raises the following warnings:

Screenshot from 2024-03-31 17-58-20

This should be an empty object, which cannot be `null`.

piotrooo:
Closes #1145

### DIFF
--- a/components/abstractions/src/main/java/com/microsoft/kiota/BaseRequestConfiguration.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/BaseRequestConfiguration.java
@@ -1,5 +1,7 @@
 package com.microsoft.kiota;
 
+import jakarta.annotation.Nonnull;
+
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


piotrooo:

components/abstractions/src/main/java/com/microsoft/kiota/BaseRequestConfiguration.java
```diff
@@ -13,9 +14,13 @@ public BaseRequestConfiguration() {
         // default empty constructor
     }
 
-    /** Request headers */
-    @Nullable public RequestHeaders headers = new RequestHeaders();
+    /**
+     * Request headers
+     */
+    public RequestHeaders headers = new RequestHeaders();
```
andrueastman, line 20: I believe the annotation should change to `@NonNull`

andrueastman (commented):
Any chance you can also add a entry to the changelog file at the root of the repositoy as well?

andrueastman (approved):
Thanks for the contribution @piotrooo

PR #1150, opened 2024-04-02
Introduce MultipartBody filename directive

### LINKED ISSUE #1146: Add `filename` directive to the `MultipartBody`
When a file is sent via an HTML form, there is an additional property in the [`Content-Disposition` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#filename) named `filename`, like this:

```
Content-Disposition: form-data; name="file"; filename="image.png"
```

It's also required when handling `multipart/form-data`, for example, in Spring. 

Currently the `MultipartBody` setting only `name`.

```java
writer.writeStringValue(
        "Content-Disposition",
        "form-data; name=\"" + originalNames.get(partEntry.getKey()) + "\"");
```

Can we conditionally add this directive?

piotrooo:
Closes #1146

### DIFF
--- a/components/abstractions/src/main/java/com/microsoft/kiota/MultipartBody.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/MultipartBody.java
@@ -48,6 +48,20 @@ public class MultipartBody implements Parsable {
      */
     public <T> void addOrReplacePart(
             @Nonnull final String name, @Nonnull final String contentType, @Nonnull final T value) {
+        addOrReplacePart(name, contentType, value, null);
+    }
+
+    /**
+     * Adds or replaces a part in the multipart body.
+     *
+     * @param <T> the type of the part to add or replace.
+     * @param name the name of the part to add or replace.
+     * @param contentType the content type of the part to add or replace.
+     * @param value the value of the part to add or replace.
+     * @param filename the value of the filename directive.
+     */
+    public <T> void addOrReplacePart(
+            @Nonnull final String name, @Nonnull final String contentType, @Nonnull final T value, @Nullable String filename) {
         Objects.requireNonNull(value);
         if (Compatibility.isBlank(contentType))
             throw new IllegalArgumentException("contentType cannot be blank or empty");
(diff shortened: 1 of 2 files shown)


andrueastman (commented):
Thanks for the contribution @piotrooo 

Any chance you can 
- Run `./gradlew spotlessApply` to fix the lint errors in the PR
- Update the changelog in the repository root

Otherwise, this looks good to me.

components/abstractions/src/main/java/com/microsoft/kiota/MultipartBody.java
```diff
@@ -48,19 +48,35 @@ public MultipartBody() {
...
         Objects.requireNonNull(value);
         if (Compatibility.isBlank(contentType))
             throw new IllegalArgumentException("contentType cannot be blank or empty");
         if (Compatibility.isBlank(name))
             throw new IllegalArgumentException("name cannot be blank or empty");
 
         final String normalizedName = normalizePartName(name);
-        originalNames.put(normalizedName, name);
-        parts.put(normalizedName, new AbstractMap.SimpleEntry<>(contentType, value));
+        Part part = new Part(name, value, contentType, filename);
```
andrueastman, line 75: This line is failing the CI with the error 
```
Error: Access to private constructor of class Part requires synthetic accessor [SyntheticAccessor]

   Explanation for issues of type "SyntheticAccessor":
   A private inner class which is accessed from the outer class will force the
   compiler to insert a synthetic accessor; this means that you are causing
   extra overhead. This is not important in small projects, but is important
   for large apps running up against the 64K method handle limit, and
   especially for libraries where you want to make sure your library is as
   small as possible for the cases where your library is used in an app
   running up against the 64K limit.

```

Does it makes sense to change to using static a factory instead for the `Part` class?

piotrooo: In this particular case, this class may not be static.

andrueastman: Just to clarify, what I mean here is that we can add a static method to the type that returns a new instance of the class rather than a private constructor?

piotrooo: @andrueastman maybe to avoid generating a synthetic constructor, we need to set it as package-scoped? That might be fair enough for Android.

andrueastman: Yeah. That could also work. 
Any chance you can make the change so that we merge this in? @piotrooo

piotrooo: Yeah, sure.

andrueastman (approved):

PR #1239, opened 2024-04-25
Normalize UUID path/query parameter values to string

Ndiritu:
closes https://github.com/microsoft/kiota-java/issues/1129


baywet (changes requested):
Can you add a unit test to prevent regressions please?

andrueastman (approved):

baywet (approved):

PR #1438, opened 2024-07-12
Group dependabot updates into single PR

Ndiritu:
- Fixes scenarios where android specific dependencies need to be bumped across each project within one PR to prevent conflicts in versions across each module during testing e.g. https://github.com/microsoft/kiota-java/actions/runs/9899563697/job/27348715110 needing the same changes made across the board to build. Linting builds all android packages together.

- Prevents duplicate PRs bumping the same dependency file e.g. #1427 and #1429 since each /android project applies the root project's dependencies & [gradle](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#gradle) considers the dependencies files in the root projects as well during updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,7 @@ updates:
     - "/components/http/okHttp"
   schedule:
     interval: daily
+    time: "09:00" # 9am UTC
   open-pull-requests-limit: 10
 - package-ecosystem: gradle
   directories:
@@ -24,6 +25,7 @@ updates:
     - "/components/http/okHttp/android"
   schedule:
     interval: daily
+    time: "10:00" # 10am UTC. Checked after the core modules to prevent duplicate PRs updating dependencies.gradle files
   open-pull-requests-limit: 10
 - package-ecosystem: github-actions
   directory: "/"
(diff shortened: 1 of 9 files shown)


baywet (approved):
Can you do the same for the Microsoft graph Java repos please?

.github/dependabot.yml
```diff
@@ -1,72 +1,14 @@
...
-  open-pull-requests-limit: 10
-- package-ecosystem: gradle
-  directory: "/components/authentication/azure/android"
-  schedule:
-    interval: daily
-  open-pull-requests-limit: 10
-- package-ecosystem: gradle
-  directory: "/components/http/okHttp/android"
+  directories:
+    - "/components/abstractions/android"
```
baywet, line 5: The plugins for the regular projects won't be updated anymore

baywet (changes requested):

.github/dependabot.yml
Ndiritu: Good callout. 
I added configs for the root project and other core modules. I staggered the checks by time since with both configs we'd still get duplicate PRs.  Staggering should prevent it.

baywet (approved):

Issue #1257, opened 2024-05-02
security - automate the OSS portal steps


baywet:
In the current releasing process, we manually go to the maven central portal to close and release the artifacts our CI uploaded.
Not only this is time consuming, it's a security concern (since this platform doesn't support organizations, we're sharing credentials)

We should explore automating those steps, and once we have the recipe, replicate on other java repositories.

https://central.sonatype.org/publish/publish-gradle/#releasing-the-deployment-to-the-central-repository
https://github.com/gradle-nexus/publish-plugin/

> Note: this wouldn't be necessary if ESRP release was available for GHA.

- [x] Remove approval steps on the maven central GitHub environment
- [x] Close and release staging repo using `closeAndReleaseSonatypeStagingRepository` after validating that new publish tasks work as expected

andrueastman:
@Ndiritu to confirm if this is possible to do in the next couple of weeks when he gets back.

PR #1445, opened 2024-07-16
Automate releases from OSSRH to Maven Central

Ndiritu:
- Adds the [Gradle Nexus Publish Plugin](https://github.com/gradle-nexus/publish-plugin/?tab=readme-ov-file#gradle-nexus-publish-plugin) which allows closing and releasing from Nexus to Maven Central.
- Unfortunately it doesn't support dropping staging repositories. We also cannot download packages from staging repositories. This PR adds validation checks to ensure certain files are part of the package contents to be uploaded. The JARs are first published to the local maven cache & file contents inspected.
- For the prod release, we are only closing the staging repository using `closeSonatypeStagingRepository`. This will allow us to inspect the staging repository manually after these changes before we can switch to `closeAndReleaseSonatypeStagingRepository` with confidence that the automation works as expected.

part of https://github.com/microsoft/kiota-java/issues/1257

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ permissions:
   contents: write
 
 env:
-  PREVIEW_TASK: 'publishToSonatype'
+  PREVIEW_TASK: publishToSonatype
   PUBLISH_TASK: publishMavenPublicationToSonatypeRepository
 
 jobs:
@@ -18,6 +18,9 @@ jobs:
     if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     environment: maven_central_snapshot
+    defaults:
+      run:
+        working-directory: ./
     steps:
     - uses: actions/checkout@v4
     - name: Setup JDK
@@ -34,14 +37,12 @@ jobs:
       env:
         ENCODED_VALUE: ${{ secrets.LOCAL_PROPERTIES }}
         OUTPUT_PATH: 'local.properties'
-      working-directory: ./
     - name: Download file
       run: .\scripts\decodeAndWrite.ps1 -encodedValue $env:ENCODED_VALUE -outputPath $env:OUTPUT_PATH
       shell: pwsh
       env:
         ENCODED_VALUE: ${{ secrets.SECRING_GPG }}
         OUTPUT_PATH: 'secring.gpg'
-      working-directory: ./
     - name: Copy secring
       run: |
         Copy-Item secring.gpg components/abstractions/ -Verbose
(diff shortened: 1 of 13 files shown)


.github/workflows/release.yml
```diff
@@ -10,19 +10,22 @@ permissions:
   contents: write
 
 env:
-  PREVIEW_TASK: publishMavenPublicationToSonatypeSnapshotRepository
-  PUBLISH_TASK: publishMavenPublicationToSonatypeRepository
+  PREVIEW_TASK: publishToSonatype
```
baywet, line 13: QQ: why did we have to change the tasks names here? (I understand why we had to add one to the publish)

baywet (commented):
LGTM besides a question

.github/workflows/release.yml
Ndiritu: I changed these to use the built-in tasks provided by the `gradle-nexus-publishing` plugin. We could still revert to the previous publish tasks provided by `maven-publish`.

However, I'd suggest we drop the maven-publish repository [configs](https://github.com/microsoft/kiota-java/blob/feat/automate-ossrh-release/components/abstractions/build.gradle#L105) and only keep the gradle-nexus-publishing ones defined in this PR so that we have only one set of repository configs. If this sounds good, I can make these changes to this PR.

This way we would only have one working set of publish tasks provided by gradle-nexus.

baywet: The reason why we had multiple different tasks was because we needed to publish to another feed in the past. That feed went offline and if we can simplify the configuration that will be awesome.

baywet: thanks for making the updates. I'm still unclear how the snapshot (preview) builds will get to the snapshot feed with the current configuration?

.github/workflows/release.yml
```diff
@@ -10,19 +10,22 @@ permissions:
...
+  PUBLISH_TASK: publishToSonatype closeSonatypeStagingRepository
```
baywet, line 14:
```suggestion
  PUBLISH_TASK: publishToSonatype closeAndReleaseSonatypeStagingRepository
```

.github/workflows/release.yml
Ndiritu: The default URLs are [pre-configured](https://github.com/gradle-nexus/publish-plugin?tab=readme-ov-file#publishing-to-maven-central-via-sonatype-ossrh)

.github/workflows/release.yml
Ndiritu: I'd be more comfortable if we didn't automate the 'release' step just yet to validate that nothing is broken with the next release. Thoughts?

baywet: oh yeah, I forgot about that comment you made earlier. disregard this one, we can automate the final step once we have vetted all of this.

.github/workflows/release.yml
Ndiritu: Ah! I get you. Good question.
It's determined by the `-SNAPSHOT` suffix being present in the version. This is currently the [default](https://github.com/microsoft/kiota-java/blob/main/gradle.properties#L41). We remove the suffix for staging to happen.

Docs weren't too clear. Found this [test](https://github.com/gradle-nexus/publish-plugin/blob/master/src/compatTest/kotlin/io/github/gradlenexus/publishplugin/BaseNexusPublishPluginTests.kt#L90)

baywet: Thank you for the additional information. In the current workflow, there's no place I could identify where we might be "removing" the snapshot suffix for central releases. Did I miss something?

Ndiritu: These were already being removed [here](https://github.com/microsoft/kiota-java/pull/1445/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R148-R162)

baywet: thanks for clarifying, I had missed this subtility

baywet (approved):

PR #1537, opened 2024-08-19
Ensure reading request body does not drain the input stream if stream is not resettable

Ndiritu:
We currently override the Request body's `writeTo` method which was not resetting the input stream after the body is written e.g. by a logging interceptor. The network call would find an empty request body.

This PR determines if the request body is resettable & configures [`isOneShot()`](https://square.github.io/okhttp/5.x/okhttp/okhttp3/-request-body/is-one-shot.html) to prevent non-resettable streams from being drained by logging middleware.

closes https://github.com/microsoftgraph/msgraph-sdk-java/issues/2037

### DIFF
--- a/components/http/okHttp/gradle/dependencies.gradle
+++ b/components/http/okHttp/gradle/dependencies.gradle
@@ -6,6 +6,7 @@ dependencies {
     // Use JUnit Jupiter Engine for testing.
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
     testImplementation 'org.mockito:mockito-inline:5.2.0'
+    testImplementation 'com.squareup.okhttp3:logging-interceptor:4.12.0'
 
 
     // This dependency is used internally, and not exposed to consumers on their own compile classpath.
--- a/components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
+++ b/components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
@@ -900,6 +900,11 @@ public class OkHttpRequestAdapter implements com.microsoft.kiota.RequestAdapter 
                                 }
 
                                 @Override
+                                public boolean isOneShot() {
+                                    return true;
+                                }
+
+                                @Override
                                 public long contentLength() throws IOException {
                                     final Set<String> contentLength =
                                             requestInfo.headers.getOrDefault(
--- a/components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
+++ b/components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
@@ -901,7 +901,7 @@ public class OkHttpRequestAdapter implements com.microsoft.kiota.RequestAdapter 
 
                                 @Override
                                 public boolean isOneShot() {
-                                    return true;
+                                    return !requestInfo.content.markSupported();
                                 }
 
                                 @Override
@@ -928,6 +928,9 @@ public class OkHttpRequestAdapter implements com.microsoft.kiota.RequestAdapter 
                                 @Override
                                 public void writeTo(@Nonnull BufferedSink sink) throws IOException {
                                     sink.writeAll(Okio.source(requestInfo.content));
+                                    if (!isOneShot()) {
+                                        requestInfo.content.reset();
+                                    }
                                 }
                             };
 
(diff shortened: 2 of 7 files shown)


components/http/okHttp/gradle/dependencies.gradle
```diff
@@ -12,6 +12,7 @@ dependencies {
     implementation 'io.opentelemetry:opentelemetry-api:1.41.0'
     implementation 'io.opentelemetry:opentelemetry-context:1.41.0'
     implementation 'jakarta.annotation:jakarta.annotation-api:2.1.1'
+    implementation 'com.google.guava:guava:33.3.0-jre'
```
baywet, line 15: some context you might not have: we explicitly removed guava from the dependencies with v6 to avoid multiple issues it was causing (https://github.com/microsoftgraph/msgraph-sdk-java/issues/1860 and others)
We should probably look at a solution that does not involve bringing back guava

baywet (changes requested):

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
```diff
@@ -923,6 +923,7 @@ public long contentLength() throws IOException {
                                 @Override
                                 public void writeTo(@Nonnull BufferedSink sink) throws IOException {
                                     sink.writeAll(Okio.source(requestInfo.content));
+                                    requestInfo.content.reset(); // unsafe
```
Ndiritu, line 926: unsafe operation. Adding for discussion purposes on alternatives allowing multiple reads from the stream without copying contents into an in-memory buffer

baywet:
for others who might be following along:
- the logging interceptor is [draining the stream](https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt#L219)
- when the transport picks it up, it's already drained
- the main reason it's most likely not doing a buffered copy is because they instead leverage the "isOneShot" method on the body to avoid draining the request body if it shouldn't.
- Our implementation needs to implement that method in the request adapter, based on whether the underlying stream is markable or not
- additionally, when the underlying stream is markable, in the writeTo method, we need to reset/mark it to zero
- This option is much better in terms of performance than buffering everything for this edge case.
- It's also much better in terms of UX than giving a setting on the request adapter to buffer ONLY when requested.

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
Ndiritu: Added condition to check if resettable, before performing reset

components/http/okHttp/src/test/java/com/microsoft/kiota/http/OkHttpRequestAdapterTest.java
```diff
@@ -401,6 +406,135 @@ void getRequestFromRequestInformationWithoutContentLengthOverrideWithEmptyPayloa
...
+        var nativeResponseHandler = new NativeResponseHandler();
+        requestInformation.setResponseHandler(nativeResponseHandler);
+
+        final var testFile = new File("./src/test/resources/helloWorld.txt");
+        final var contentLength = testFile.length();
+
+        try (FileInputStream content = new FileInputStream(testFile)) {
+            requestInformation.setStreamContent(content, "application/octet-stream");
+            var mockEntity = creatMockEntity();
+            requestAdapter.send(requestInformation, null, (node) -> mockEntity);
```
baywet, line 531:
```suggestion
            requestAdapter.send(requestInformation, null, node -> mockEntity);
```

components/http/okHttp/src/test/java/com/microsoft/kiota/http/OkHttpRequestAdapterTest.java
```diff
@@ -401,6 +406,135 @@ void getRequestFromRequestInformationWithoutContentLengthOverrideWithEmptyPayloa
...
+        var requestBodyJson = "{\"name\":\"value\",\"array\":[\"1\",\"2\",\"3\"]}";
+        ByteArrayInputStream content =
+                new ByteArrayInputStream(requestBodyJson.getBytes(StandardCharsets.UTF_8));
+        requestInformation.setStreamContent(content, "application/json");
+        requestInformation.httpMethod = HttpMethod.PUT;
+        var nativeResponseHandler = new NativeResponseHandler();
+        requestInformation.setResponseHandler(nativeResponseHandler);
+
+        var mockEntity = creatMockEntity();
+        requestAdapter.send(requestInformation, null, (node) -> mockEntity);
```
baywet, line 496:
```suggestion
        requestAdapter.send(requestInformation, null, node -> mockEntity);
```

baywet (dismissed):

Ndiritu:
> Thank you for making the changes! A version bump, a changelog entry and we should be good to go!

Ah no release please yet :)
On it.

baywet (approved):

baywet:
@Ndiritu the tag also needs to be manually created on this repo for the time being. I'll let you take care of that :)

mkomko:
@Ndiritu @baywet Thanks a lot guys! ❤️

PR #1538, opened 2024-08-22
Fix Java 21 Build issues

Ndiritu:
Fixes failing builds in [release](https://github.com/microsoft/kiota-java/actions/runs/10509355140/job/29115355499) workflow

- Fixes failing tests caused by our previous dependency on `mockito-inline` which transitively installed [`byte-buddy-agent:1.14.1`](https://central.sonatype.com/artifact/net.bytebuddy/byte-buddy-agent/1.14.1). This version of byte-buddy agent [doesn't support Java 21](https://github.com/raphw/byte-buddy/issues/1396#issuecomment-1763717640). `mockito-inline` was deprecated after `5.3.0` of `mockito-core`. Changing to `mockito-core`'s latest version allows us to transitively get a Java 21 compatible `byte-buddy` dependency.
- Updates workflows to use Java 21

Pending:
- Change serialization validation [here](https://github.com/microsoft/kiota-java/blob/main/components/serialization/json/src/test/java/com/microsoft/kiota/serialization/JsonSerializationWriterTests.java#L191) since order of set entries is not guaranteed during [serializing](https://github.com/microsoft/kiota-java/blob/main/components/serialization/json/src/main/java/com/microsoft/kiota/serialization/JsonSerializationWriter.java#L359). Java 17 & 21 produce different orderings.

### DIFF
--- a/.github/workflows/api-level-lint.yml
+++ b/.github/workflows/api-level-lint.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
-          java-version: 17
+          java-version: 21
           cache: gradle
       - name: Setup Android SDK
         uses: android-actions/setup-android@v3.2.1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Setup JDK
       uses: actions/setup-java@v4
       with:
-        java-version: 17
+        java-version: 21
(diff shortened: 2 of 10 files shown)


baywet:
@Ndiritu can you create issues in the microsoft graph repositories to replace the [deprecated dependency please](https://github.com/search?q=org%3Amicrosoftgraph+mockito-inline+language%3AGradle&type=code&l=Gradle)?

baywet (approved):

PR #1558, opened 2024-09-06
Fix undefined behaviour in backing store

Ndiritu:
Fixes a bug encountered while marking all properties in an `InMemoryBackingStore` as dirty.

### Context

When iterating over a `Map` [`entrySet()`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Map.html#entrySet()), any modification to the underlying map causes undefined behavior during the iteration.

The `InMemoryBackingStore` iterates over the underlying store's [`entrySet()`](https://github.com/microsoft/kiota-java/blob/main/components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java#L57) and currently calls [`ensureCollectionPropertyIsConsistent()`](https://github.com/microsoft/kiota-java/blob/main/components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java#L65C13-L65C49) which may update the underlying Map by calling [`set()`](https://github.com/microsoft/kiota-java/blob/main/components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java#L231) to update the collection size & flag the collection as dirty in case of additions/deletions. 

Undefined behaviour manifests as repeated iterations over the same set of items in the underlying store causing performance issues. The situation is amplified when we have a collection of `BackedModel`s where each `BackedModel` has a collection property that has been updated and we need to mark all properties of all the `BackedModel`s as dirty. 

### Changes

This PR:
- Updates check for collection size consistency to prevent updating the underlying map during iteration.
- Reduces redundant recursive calls when cascading dirty tracking to nested `BackedModel` elements
- Separates concerns by allowing `get()` to determine whether a value is dirty or not which matches its interface [definition](https://github.com/microsoft/kiota-java/blob/main/components/abstractions/src/main/java/com/microsoft/kiota/store/BackingStore.java#L15). `getValueFromWrapper()` now only unwraps values from the underlying store
- Fixes bug to ensure collection properties are checked for consistency before determining if they are dirty when we want to return only changed values

The sample test case would previously fail after running for `~60s` vs with this fix, it passes in `~3s` on the same hardware.

closes https://github.com/microsoftgraph/msgraph-sdk-java/issues/2106

### DIFF
--- a/components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
@@ -80,7 +80,8 @@ public class InMemoryBackingStore implements BackingStore {
                             BackedModel backedModel = (BackedModel) item;
                             backedModel
                                     .getBackingStore()
-                                    .setIsInitializationCompleted(value); // propagate initialization
+                                    .setIsInitializationCompleted(
+                                            value); // propagate initialization
                         }
                     }
                 }
@@ -112,9 +113,7 @@ public class InMemoryBackingStore implements BackingStore {
                     for (final Object item : items) {
                         if (item instanceof BackedModel) {
                             BackedModel backedModel = (BackedModel) item;
-                            backedModel
-                                    .getBackingStore()
-                                    .setReturnOnlyChangedValues(value);
+                            backedModel.getBackingStore().setReturnOnlyChangedValues(value);
                         }
                     }
                 }
--- a/components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
@@ -235,8 +235,8 @@ public class InMemoryBackingStore implements BackingStore {
     }
 
     private void ensureCollectionPropertiesAreConsistent() {
-        HashMap<String, Object> currentStoreDirtyCollections = new HashMap<>();
-        List<BackedModel> nestedBackedModelsToEnumerate = new ArrayList<>();
+        final HashMap<String, Object> currentStoreDirtyCollections = new HashMap<>();
+        final List<BackedModel> nestedBackedModelsToEnumerate = new ArrayList<>();
 
         for (final Map.Entry<String, Pair<Boolean, Object>> entry : this.store.entrySet()) {
             final Pair<Boolean, Object> wrapper = entry.getValue();
@@ -260,7 +260,7 @@ public class InMemoryBackingStore implements BackingStore {
         }
 
         // Only update parent properties that haven't been marked as dirty by the nested models
-        for (Map.Entry<String, Object> entry : currentStoreDirtyCollections.entrySet()) {
+        for (final Map.Entry<String, Object> entry : currentStoreDirtyCollections.entrySet()) {
             // Always set() if there were no nested models
             if (nestedBackedModelsToEnumerate.isEmpty()) {
                 set(entry.getKey(), entry.getValue());
--- a/components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
@@ -54,9 +54,11 @@ public class InMemoryBackingStore implements BackingStore {
 
     public void setIsInitializationCompleted(final boolean value) {
         this.isInitializationCompleted = value;
+        ensureCollectionPropertiesAreConsistent();
         for (final Map.Entry<String, Pair<Boolean, Object>> entry : this.store.entrySet()) {
             final Pair<Boolean, Object> wrapper = entry.getValue();
             Pair<Boolean, Object> updatedValue = new Pair<>(!value, wrapper.getValue1());
+            entry.setValue(updatedValue);
 
             if (wrapper.getValue1() instanceof BackedModel) {
                 BackedModel backedModel = (BackedModel) wrapper.getValue1();
@@ -65,28 +67,18 @@ public class InMemoryBackingStore implements BackingStore {
                         .setIsInitializationCompleted(value); // propagate initialization
             }
             if (wrapper.getValue1() instanceof Pair) {
-                // setIsInitializationCompleted() called above already checks for collection
-                // consistency for BackedModels
                 final Pair<?, Integer> collectionTuple = (Pair<?, Integer>) wrapper.getValue1();
-                Object[] items;
-                if (collectionTuple.getValue0() instanceof Collection) {
-                    items = ((Collection<Object>) collectionTuple.getValue0()).toArray();
-                } else { // it is a map
-                    items = ((Map<?, Object>) collectionTuple.getValue0()).values().toArray();
-                }
+                Object[] items = getObjectArrayFromCollectionWrapper(collectionTuple);
 
                 for (final Object item : items) {
-                    touchNestedProperties(item); // call get on nested properties
-                }
+                    if (!(item instanceof BackedModel)) break;
 
-                if (collectionTuple.getValue1()
-                        != items.length) { // and the size has changed since we last updated
-                    updatedValue =
-                            new Pair<>(
-                                    !value, new Pair<>(collectionTuple.getValue0(), items.length));
+                    BackedModel backedModel = (BackedModel) item;
+                    backedModel
+                            .getBackingStore()
+                            .setIsInitializationCompleted(value); // propagate initialization
                 }
             }
-            entry.setValue(updatedValue);
         }
     }
 
@@ -155,7 +147,7 @@ public class InMemoryBackingStore implements BackingStore {
 
         boolean hasChanged = wrapper.getValue0();
         if (getReturnOnlyChangedValues() && !hasChanged) {
-            ensureCollectionPropertyIsConsistent(key, wrapper.getValue1());
+            ensureCollectionPropertiesAreConsistent();
             hasChanged = this.store.get(key).getValue0();
             if (!hasChanged) {
                 return null;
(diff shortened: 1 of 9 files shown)


baywet:
You might want to look at the follow PRs and see if any of the fixes/optimizations could be ported here.
https://github.com/microsoft/kiota-dotnet/pull/95
https://github.com/microsoft/kiota-dotnet/pull/311
https://github.com/microsoft/kiota-dotnet/pull/326

Ndiritu:
> You might want to look at the follow PRs and see if any of the fixes/optimizations could be ported here. [microsoft/kiota-dotnet#95](https://github.com/microsoft/kiota-dotnet/pull/95) [microsoft/kiota-dotnet#311](https://github.com/microsoft/kiota-dotnet/pull/311) [microsoft/kiota-dotnet#326](https://github.com/microsoft/kiota-dotnet/pull/326)

Missing KiotaSerialization helper updates. To be tracked by https://github.com/microsoft/kiota-java/issues/1561 next sprint

components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
```diff
@@ -48,24 +49,37 @@ public Pair<A, B> setValue1(B value1) {
...
     private final Map<String, TriConsumer<String, Object, Object>> subscriptionStore =
-            new HashMap<>();
+            new ConcurrentHashMap<>();
 
     public void setIsInitializationCompleted(final boolean value) {
         this.isInitializationCompleted = value;
+        ensureCollectionPropertiesAreConsistent();
         for (final Map.Entry<String, Pair<Boolean, Object>> entry : this.store.entrySet()) {
             final Pair<Boolean, Object> wrapper = entry.getValue();
+            Pair<Boolean, Object> updatedValue = new Pair<>(!value, wrapper.getValue1());
```
baywet, line 61:
```suggestion
            final Pair<Boolean, Object> updatedValue = new Pair<>(!value, wrapper.getValue1());
```

components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
```diff
@@ -112,20 +126,13 @@ public void clear() {
...
-                if (Boolean.FALSE.equals(
-                        hasChanged)) { // no need property has already been flagged.
-                    ensureCollectionPropertyIsConsistent(entryKey, wrapper.getValue1());
-                }
-                if (wrapper.getValue1() instanceof Pair) {
-                    Pair<?, ?> collectionTuple = (Pair<?, ?>) wrapper.getValue1();
-                    return collectionTuple.getValue0();
-                }
-                return wrapper.getValue1();
+            if (wrapper.getValue1() instanceof Pair) {
```
shemogumbe, line 131: We could do an early return, like a null check followed by an early return, will be more readable and reduce nesting

```java
 if (wrapper == null) {
        return null;
    }
```

shemogumbe: And maybe a javadoc comment

baywet, line 131: I'm struggling to follow why the value would be a Pair? In which case would we have nested pairs?

components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
```diff
@@ -212,39 +232,50 @@ private void setupNestedSubscriptions(
...
-        if (nestedObject instanceof BackedModel) {
-            // Call Get<>() on nested properties so that this method may be called recursively to
-            // ensure collections are consistent
-            final BackedModel backedModel = (BackedModel) nestedObject;
-            for (final String itemKey : backedModel.getBackingStore().enumerate().keySet()) {
-                backedModel.getBackingStore().get(itemKey);
-            }
+    private Object[] getObjectArrayFromCollectionWrapper(final Pair<?, Integer> collectionTuple) {
+        if (collectionTuple.getValue0() instanceof Collection) {
+            return ((Collection<Object>) collectionTuple.getValue0()).toArray();
```
baywet, line 276: assign get value0 to a variable

components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
```diff
@@ -212,39 +232,50 @@ private void setupNestedSubscriptions(
...
+        } else { // it is a map
```
baywet, line 277: could we test it so we don't end up throwing a cast exception for a case we didn't plan for?

components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
```diff
@@ -134,7 +141,20 @@ private Object getValueFromWrapper(final String entryKey, final Pair<Boolean, Ob
...
         final Pair<Boolean, Object> wrapper = this.store.get(key);
-        final Object value = this.getValueFromWrapper(key, wrapper);
+        if (wrapper == null) {
+            return null;
+        }
+        final Object value = this.getValueFromWrapper(wrapper);
+
+        boolean hasChanged = wrapper.getValue0();
+        if (getReturnOnlyChangedValues() && !hasChanged) {
+            ensureCollectionPropertiesAreConsistent();
```
baywet, line 151: I'm worried about the performance of enumerating everything on every get. Can you please provide more context on why this is required here?

baywet (changes requested):

components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
```diff
@@ -212,39 +232,50 @@ private void setupNestedSubscriptions(
...
+                }
+                if (collectionTuple.getValue1()
+                        != items.length) { // and the size has changed since we last updated
+                    currentStoreDirtyCollections.put(entry.getKey(), collectionTuple.getValue0());
+                }
             }
+        }
+
+        // Enumerate nested backed models first since they may trigger the parent to be dirty
+        for (BackedModel nestedBackedModel : nestedBackedModelsToEnumerate) {
```
shemogumbe, line 256: We could do some error handling for cases where collections are not an instance of BackedModel

Ndiritu: If the collection doesn't contain `BackedModel`s, `nestedBackedModelsToEnumerate` will be empty:
https://github.com/microsoft/kiota-java/pull/1558/files#diff-d7e6365ce63b26ccd48ddba99ccc90260d769259ee24bd1987812fd72be950a8R245

components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
Ndiritu: Will add a helper method for these checks to be clearer.

components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
Ndiritu: If a nested collection is updated directly, the backing store has no way of being notified of this change e.g.
```java
backedModel.getAdditionalData().put("newKey", "newVal");
```
We need to run through the hierarchy of `BackedModel`s looking for any collection values & check their sizes.

Yes, this is expensive but:
- These checks are limited to collection properties that are NOT dirty since they may have gotten an addition/removal directly. 
- This check should happen at most once on each `BackedModel` in a hierarchy. Once a collection property is marked as dirty by these checks, future get()'s on the same `BackedModel` should skip this.

I can further limit this by doing this only if the value for the `key` is a collection.

baywet: thanks for providing more information here. Yes, let's try to reduce iterations as much as possible.

components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
```diff
@@ -212,39 +234,61 @@ private void setupNestedSubscriptions(
...
-            final Pair<?, Integer> collectionTuple = (Pair<?, Integer>) storeItem;
-            Object[] items;
-            if (collectionTuple.getValue0() instanceof Collection) {
-                items = ((Collection<Object>) collectionTuple.getValue0()).toArray();
-            } else { // it is a map
-                items = ((Map<?, Object>) collectionTuple.getValue0()).values().toArray();
-            }
+    private void ensureCollectionPropertiesAreConsistent() {
+        HashMap<String, Object> currentStoreDirtyCollections = new HashMap<>();
+        List<BackedModel> nestedBackedModelsToEnumerate = new ArrayList<>();
```
baywet, line 239:
```suggestion
        final HashMap<String, Object> currentStoreDirtyCollections = new HashMap<>();
        final List<BackedModel> nestedBackedModelsToEnumerate = new ArrayList<>();
```

components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
```diff
@@ -212,39 +234,61 @@ private void setupNestedSubscriptions(
...
-                        key,
-                        collectionTuple.getValue0()); // ensure the store is notified the collection
-                // property is "dirty"
+        // Enumerate nested backed models first since they may trigger the parent to be dirty
+        for (BackedModel nestedBackedModel : nestedBackedModelsToEnumerate) {
+            nestedBackedModel.getBackingStore().enumerate();
+        }
+
+        // Only update parent properties that haven't been marked as dirty by the nested models
+        for (Map.Entry<String, Object> entry : currentStoreDirtyCollections.entrySet()) {
```
baywet, line 263:
```suggestion
        for (final Map.Entry<String, Object> entry : currentStoreDirtyCollections.entrySet()) {
```

components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
```diff
@@ -212,39 +234,61 @@ private void setupNestedSubscriptions(
...
+        final List<BackedModel> nestedBackedModelsToEnumerate = new ArrayList<>();
 
-            for (final Object item : items) {
-                touchNestedProperties(item); // call get on nested properties
+        for (final Map.Entry<String, Pair<Boolean, Object>> entry : this.store.entrySet()) {
+            final Pair<Boolean, Object> wrapper = entry.getValue();
+            if (isCollectionValue(wrapper)) {
+                final Pair<?, Integer> collectionTuple = (Pair<?, Integer>) wrapper.getValue1();
+                Object[] items = getObjectArrayFromCollectionWrapper(collectionTuple);
+                for (Object item : items) {
```
baywet, line 246:
```suggestion
                for (final Object item : items) {
```

baywet (approved):
I'd like @andrueastman revie as well before we merge this one

components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
```diff
@@ -212,39 +234,61 @@ private void setupNestedSubscriptions(
...
-            for (final Object item : items) {
-                touchNestedProperties(item); // call get on nested properties
+        for (final Map.Entry<String, Pair<Boolean, Object>> entry : this.store.entrySet()) {
+            final Pair<Boolean, Object> wrapper = entry.getValue();
+            if (isCollectionValue(wrapper)) {
+                final Pair<?, Integer> collectionTuple = (Pair<?, Integer>) wrapper.getValue1();
+                Object[] items = getObjectArrayFromCollectionWrapper(collectionTuple);
+                for (final Object item : items) {
+                    if (!(item instanceof BackedModel)) break;
+                    nestedBackedModelsToEnumerate.add((BackedModel) item);
```
andrueastman, line 248: Similar concern here, the cast will most likely fail if we called `getObjectArrayFromCollectionWrapper` on a map that has mixed types and the first item happened to be a backed model.

andrueastman (commented):
Just a couple of comments from me. Otherwise this looks good to me. 
Thanks for looking into this @Ndiritu

components/abstractions/src/main/java/com/microsoft/kiota/store/InMemoryBackingStore.java
Ndiritu: Thanks for catching this. Made these updates & added tests.

Tests also exposed a bug where we now need to propagate `returnOnlyChangedValues` to nested `BackedModels` to ensure nested models with collection properties are checked. We have limited the collection checks to happen during enumeration only when we want to return changed values.

andrueastman (approved):
👍🏼

PR #1586, opened 2024-09-27
Implement support for LocalDate and OffsetDateTime in RequestInformation

### LINKED ISSUE #1583: prepare for std uri template 2 and date time drop
related https://github.com/std-uritemplate/std-uritemplate/pull/245
The following types need to the normalized to their RFC3339 representation before they are passed to std URI template when they are present are query or path parameter values
Date, OffsetDateTime their collection representations (array, List)

espenvis:
Implements behavior for `LocalDate` and `OffsetDateTime` in RequestInformation's `getSanitizedValues`. Added accompanying test for `LocalDate` path parameter resolution.

Edit: scope is extended to include `LocalTime` and `PeriodAndDuration`.

closes #1583


espenvis:
Marking as draft until we can evaluate extending this issue to tackle `time` and `duration` amongst other types.

espenvis:
@microsoft-github-policy-service agree company="VisBook AS"

components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
```diff
@@ -473,6 +474,10 @@ private static Object getSanitizedValues(Object value) {
             return ((ValuedEnum) value).getValue();
         } else if (value instanceof UUID) {
             return value.toString();
+        } else if (value instanceof OffsetDateTime) {
+            return ((OffsetDateTime) value).format(DateTimeFormatter.ISO_OFFSET_DATE_TIME);
+        } else if (value instanceof LocalDate) {
```
baywet, line 479: here we'd be missing LocalTime and PeriodAndDuration

baywet (changes requested):
Thanks for the contribution!
Could you also bump the dependency now that it's been released please?
(this way we can check there are no regressions)

components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
espenvis: Yes, I will implement this now that it's been confirmed we want to support these.

espenvis:
Updated to include `time` and `duration` datatype support (`LocalTime` and `PeriodAndDuration`). Also merged upstream changes. The version is now bumped up to `1.5.0` and all tests pass. Let me know if I've missed something.

components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
espenvis: These are now implemented.

baywet (commented):
Can you also bump std uri template to v2 so we can confirm there's no end to end regression please?
https://github.com/microsoft/kiota-java/blob/bf66598153485fea2442dcf108d00e4ff6b8ab30/components/abstractions/gradle/dependencies.gradle#L11

baywet (dismissed):

espenvis:
> Thank you for making the changes! Can you also bump std uri template to v2 so we can confirm there's no end to end regression please?
> 
> https://github.com/microsoft/kiota-java/blob/bf66598153485fea2442dcf108d00e4ff6b8ab30/components/abstractions/gradle/dependencies.gradle#L11

Yes. Apologies, I missed your earlier question.

It seems this causes an exception in `UriTemplateTest.java:32` - this passes parameters directly to `StdUriTemplate` and will fail.

baywet:
@espenvis thanks for the information.
I think you can remove the last line in the CSV and the date parameter in the unit test. (make sure you pull beforehand since I've applied formatting to the csv file)

espenvis:
> @espenvis thanks for the information. I think you can remove the last line in the CSV and the date parameter in the unit test. (make sure you pull beforehand since I've applied formatting to the csv file)

Yep, that's what I figured. I'll do that.

espenvis:
Test still fails due to a JUnit error of some sort. I'll have a look.

espenvis:
Seems it complained about the change in delimiter. Pushing a fix for that.

baywet (approved):

baywet (approved):

Issue #1498, opened 2024-08-13
Add release-please configuration


Ndiritu:
We want:
 - [ ] release please header file (yaml .github/release-please.yml)
 - [ ] release please manifest (json .release-please-manifest.json)
 - [ ] release please config (json release-please-config.json)
 - [ ] changelog keep a changelog header and unreleased information to be removed
 - [ ] contributing to be updated with commit formatting instructions
 - [ ] delete any version bump scripts
 - [ ] files where the version needs to be updated to be updated with the special comments for release please

PR #1605, opened 2024-10-02
Adds release please configuration

Ndiritu:
closes https://github.com/microsoft/kiota-java/issues/1498


baywet (approved):
LGTM other than the fact that spotless is not happy about some release please comments

release-please-config.json
```diff
@@ -0,0 +1,26 @@
...
+        ".gradle",
+        ".vscode",
+        "scripts",
+        "gradle"
+    ],
+    "release-type": "simple",
+    "include-component-in-tag": false,
+    "include-v-in-tag": true,
+    "packages": {
+        ".": {
```
calebkiage, line 16: I think you should make use of multiple packages for each library in case one of the libraries needs to release a patch version

Ndiritu:
> LGTM other than the fact that spotless is not happy about some release please comments

Let me try and resolve them + update the secrets scanning baseline

README.md
```diff
@@ -21,14 +21,29 @@ Read more about Kiota [here](https://github.com/microsoft/kiota/blob/main/README
 In `build.gradle` in the `dependencies` section:
 
 ```Groovy
+// x-release-please-start-version
```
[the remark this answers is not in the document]
baywet: yes since we always release all the packages at the same time.

release-please-config.json
baywet: the releasing is setup in such a way today that we cannot release any package independently. We'd need to update the CD to support such a scenario.

Ndiritu: We intend to bump all the packages when one changes to keep a consistent version across the board. Reduces confusion about what version of a kiota package is compatible with what etc.

README.md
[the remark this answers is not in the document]
Ndiritu: Yes

release-please-config.json
Ndiritu: @baywet thoughts about creating multiple GitHub releases for each package?

Would mean we include the component name in the tags e.g. `abstractions-v1.6.0` to prevent conflicts when we update each package's GitHub release with its relevant JAR.

Ndiritu: (Ah, forgot to refresh :). Deleting duplicate comments)

baywet: it's not just about the GH release, it's also about the upload to maven central. We'd need to somehow detect which packages have been bumped, and only perform the upload for those. I'm not saying it's impossible, just more work.
And we've already gotten the feedback that:
1. misaligned versions are confusions
2. updating all dependencies is not the initial burden we thought, especially when versions are aligned

And we've invested in the bundle package.

This is why I'm suggesting we setup RP in this PR (initial scope) and avoid expanding the scope to change a bunch of things in the versioning, CD, etc... At least until we get further feedback that we're producing too many releases.

baywet (approved):

calebkiage (approved):

Issue #1131, opened 2024-03-20
Serialization methods overloads to disable the "only changed properties behaviour"


baywet:
We recently got very similar questions:
https://github.com/microsoftgraph/msgraph-sdk-java/issues/1886
https://github.com/microsoftgraph/msgraph-sdk-java/issues/1879

People generally want to:
1. get something from the service
2. serialize it to store it locally, in a document db, etc...

While our serialization helpers are helpful, they fall short of one aspect: they don't allow the caller to specify "serialize everything" (as opposed to serializing only the changed properties).

My suggestion here is to:
1. add an overload that'd accept a boolean parameter "serializeOnlyCHangedValues" default false
2. have the existing implementations call into that overload, with default false (the behaviour will change)
3. add a routine that crawls the object tree whenever the value is false, and whenever it finds a backed model, flips the serialize only changed values from the backing store.
4. once the serialization is done, crawl the tree again to restore things.

@sebastienlevert to provide input on this one before we start any work, and if we get an agreement here, I'll replicate the issue to the other repos.

sebastienlevert:
Just so I understand. When serializing models (across languages) we only serialize what has changed when the backing store is enabled, right? If this is the case, I agree we should have options to "deep" serialize vs only what is in the backing store.

If it's the case without the backing store, then I think this is a bug, though...

baywet:
this is only present with the backing store. And we don't want to give people the option, but rather have a consistent behaviour.
When they change a collection (set the collection again), we should consider all items within the collection changed, regardless of where they came from. (this is broken)
If they only insert/remove items to the collection, then only individual items should be considered changed. (that aspect works today)

sebastienlevert:
Just to make it clear, can you provide a tiny example of before / after of a collection that needs to be serialized?

PR #1608, opened 2024-10-04
Allow backing store serialization configuration in serialization helpers

Ndiritu:
closes https://github.com/microsoft/kiota-java/issues/1131

Allows overriding default backing store behaviour while serializing.
- Retains previous serialization helper method implementation with the default behaviour of returning only changed values (`true`). To prevent backward incompatibility.

### Usage
Before:
```java
// BackedModel model = kiotaClient...get();
String result = KiotaJsonSerialization.serializeAsString(model); // returns empty string since no property has changed.
```
Now:

```java
// User model = kiotaClient...get();
String result = KiotaJsonSerialization.serializeAsString(model, false); // returns serialized string by disabling returnOnlyChangedValues
```

### DIFF
--- a/components/abstractions/src/main/java/com/microsoft/kiota/serialization/KiotaSerialization.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/serialization/KiotaSerialization.java
@@ -30,7 +30,7 @@ public final class KiotaSerialization {
      */
     @Nonnull public static <T extends Parsable> InputStream serializeAsStream(
             @Nonnull final String contentType, @Nonnull final T value) throws IOException {
-        return serializeAsStream(contentType, value, DEFAULT_SERIALIZE_ONLY_CHANGED_VALUES);
+        return serializeAsStream(contentType, DEFAULT_SERIALIZE_ONLY_CHANGED_VALUES, value);
     }
 
     /**
@@ -44,8 +44,8 @@ public final class KiotaSerialization {
      */
     @Nonnull public static <T extends Parsable> InputStream serializeAsStream(
             @Nonnull final String contentType,
-            @Nonnull final T value,
-            final boolean serializeOnlyChangedValues)
+            final boolean serializeOnlyChangedValues,
+            @Nonnull final T value)
             throws IOException {
         Objects.requireNonNull(value);
         try (final SerializationWriter writer =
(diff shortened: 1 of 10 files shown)


components/abstractions/src/main/java/com/microsoft/kiota/serialization/KiotaSerialization.java
```diff
@@ -29,7 +30,26 @@ private KiotaSerialization() {}
...
+     * @param <T> the type of the value to serialize
+     * @param contentType the content type to use for serialization
+     * @param value the value to serialize
+     * @param serializeOnlyChangedValues whether to serialize all values in value if value is a BackedModel
+     * @return the serialized value as a stream
+     * @throws IOException when the stream cannot be closed or read.
+     */
+    @Nonnull public static <T extends Parsable> InputStream serializeAsStream(
+            @Nonnull final String contentType,
+            final boolean serializeOnlyChangedValues,
```
andrueastman, line 47: should we make the boolean parameter last so that if a refator is done from the previous method, it will seem like a new parameter is added at the end of the argument list rather than at the middle?

components/abstractions/src/main/java/com/microsoft/kiota/serialization/SerializationWriterProxyFactory.java
```diff
@@ -13,7 +13,7 @@ public abstract class SerializationWriterProxyFactory implements SerializationWr
         return _concrete.getValidContentType();
     }
 
-    private final SerializationWriterFactory _concrete;
+    protected final SerializationWriterFactory _concrete;
```
baywet, line 16: We should update the field name to follow naming conventions since it's not private anymore.

components/abstractions/src/main/java/com/microsoft/kiota/serialization/SerializationWriterFactoryRegistry.java
```diff
@@ -36,20 +38,92 @@ public SerializationWriterFactoryRegistry() {
         if (contentType.isEmpty()) {
             throw new NullPointerException("contentType cannot be empty");
         }
-        final String vendorSpecificContentType = contentType.split(";")[0];
+        final ContentTypeWrapper contentTypeWrapper = new ContentTypeWrapper(contentType);
```
baywet, line 41: While I appreciate the effort to refactor things, this is going to introduce new allocations on a method that's on the hot path.

Couldn't this method simply call into the new method with a default value to avoid duplication?

baywet (changes requested):

components/abstractions/src/main/java/com/microsoft/kiota/serialization/KiotaSerialization.java
Ndiritu: Good point, followed the linting warning a little blindly. Adding a linting option to disable it since it doesn't apply in this case.

andrueastman (approved):

baywet (approved):

PR #1609, opened 2024-10-04
Fix default handler overide , when user passed in interceptor should …

raghucha:
This Pull request allows us to overide the default implementations of the interceptor behaviour . This is linked to this [bug](https://github.com/microsoftgraph/msgraph-sdk-java-core/pull/1758)


raghucha:
@baywet @Ndiritu  can you please review this , i have added a new default interceptor to be in sync with .Net code , [Net pull Request](https://github.com/microsoft/kiota-http-dotnet/pull/246)

components/http/okHttp/src/main/java/com/microsoft/kiota/http/KiotaClientFactory.java
```diff
@@ -84,6 +87,7 @@ private KiotaClientFactory() {}
      */
     @Nonnull public static Interceptor[] createDefaultInterceptors() {
         return new Interceptor[] {
+            new UrlReplaceHandler(),
```
Ndiritu, line 90: This doesn't perform any action when configured by default. I think we could leave it out and allow developers to add it when they need to configure it to perform a replacement

components/http/okHttp/src/main/java/com/microsoft/kiota/http/KiotaClientFactory.java
```diff
@@ -92,11 +96,70 @@ private KiotaClientFactory() {}
         };
     }
 
+    /**
+     * Creates the default interceptors for the client.
+     * @return an array of interceptors.
+     */
+    @Nonnull public static Interceptor[] createOveridableDefaultInterceptors( @Nonnull Interceptor[] interceptors) {
```
Ndiritu, line 103: Could still lead to unexpected behaviour since we're always adding the default middleware to the end of the chain. This was a bug in the previous implementation.

Ndiritu (commented):
Thank you for your time on this @raghucha 

I think a cleaner override workflow would be, a developer gets the default Interceptor list... and can re-arrange it as they please and when they pass it as a parameter, we add the interceptors as is to the OkHttpClient builder:
```java

var interceptors = createDefaultInterceptors();
// dev manipulates interceptors as needed, pre-pend/append custom Interceptors, remove, re-arrange etc...
KiotaClientFactory.create(interceptors); // doesn't manipulate interceptors but adds to the builder directly
```
Or they pass custom configuration options for the existing default Interceptors by adding a 
```java 
create(RequestOption[] options)
``` 
overload, which we can add in this PR. We can then match the different options and pass them to the corresponding interceptors.

Happy to help where needed.

components/http/okHttp/src/main/java/com/microsoft/kiota/http/KiotaClientFactory.java
Ndiritu: .NET only uses the custom handlers when provided. It doesn't append default handlers: [ref](https://github.com/microsoft/kiota-dotnet/blob/main/src/http/httpClient/KiotaClientFactory.cs#L54)

components/http/okHttp/src/main/java/com/microsoft/kiota/http/KiotaClientFactory.java
```diff
@@ -92,11 +102,60 @@ private KiotaClientFactory() {}
         };
     }
 
+    /**
+     * Creates the default interceptors for the client.
+     * @return an array of interceptors.
+     */
+    @Nonnull public static Interceptor[] createDefaultInterceptors( @Nullable Interceptor[] interceptors) {
```
baywet, line 109: what's the value of this method? couldn't we rely on the existing one or the one that accepts options instead to avoid a sprawling public API surface?

components/http/okHttp/src/main/java/com/microsoft/kiota/http/KiotaClientFactory.java
```diff
@@ -92,11 +102,60 @@ private KiotaClientFactory() {}
...
+        if(interceptors == null || interceptors.length == 0){
+            return createDefaultInterceptors();
+        }
+        return interceptors;
+
+    }
+
+    @Nonnull public static Interceptor[] createDefaultInterceptors(List<RequestOption> requestOptions) {
```
baywet, line 117:
```suggestion
    @Nonnull public static Interceptor[] createDefaultInterceptors(@Nullable final List<RequestOption> requestOptions) {
```

components/http/okHttp/src/main/java/com/microsoft/kiota/http/KiotaClientFactory.java
```diff
@@ -92,11 +102,60 @@ private KiotaClientFactory() {}
...
+
+        UrlReplaceHandlerOption uriReplacementOption = null;
```
baywet, line 119: default the list value if it's null

components/http/okHttp/src/main/java/com/microsoft/kiota/http/KiotaClientFactory.java
```diff
@@ -92,11 +102,60 @@ private KiotaClientFactory() {}
...
+        UserAgentHandlerOption userAgentHandlerOption = null;
+        RetryHandlerOption retryHandlerOption = null;
+        RedirectHandlerOption redirectHandlerOption = null;
+        ParametersNameDecodingOption parametersNameDecodingOption = null;
+        HeadersInspectionOption headersInspectionHandlerOption = null;
+
+        for (RequestOption option : requestOptions) {
```
baywet, line 126:
```suggestion
        for (final RequestOption option : requestOptions) {
```

components/http/okHttp/src/main/java/com/microsoft/kiota/http/KiotaClientFactory.java
baywet: this method should call into the one that accepts options and pass the options to avoid duplicated definitions.

components/http/okHttp/src/main/java/com/microsoft/kiota/http/KiotaClientFactory.java
```diff
@@ -92,11 +102,60 @@ private KiotaClientFactory() {}
...
+            } else if (parametersNameDecodingOption == null && option instanceof ParametersNameDecodingOption) {
+                parametersNameDecodingOption = (ParametersNameDecodingOption) option;
+            } else if (userAgentHandlerOption == null && option instanceof UserAgentHandlerOption) {
+                userAgentHandlerOption = (UserAgentHandlerOption) option;
+            } else if (headersInspectionHandlerOption == null && option instanceof HeadersInspectionOption) {
+                headersInspectionHandlerOption = (HeadersInspectionOption) option;
+            }
+        }
+
+        List<Interceptor> handlers = new ArrayList<>();
```
baywet, line 142:
```suggestion
        final List<Interceptor> handlers = new ArrayList<>();
```

baywet (changes requested):
Thanks for the contribution!

components/http/okHttp/src/main/java/com/microsoft/kiota/http/KiotaClientFactory.java
baywet: I'm not sure I understand the comment here? We already had a createDefaultInterceptors method (no parameters) and we're not taking it away.
We're introducing an overload which accepts a list of options.
How is this method, which accept an array of options preventing any source breaking change?
I feel like we could achieve the same result without this specific method.

raghucha: @baywet  just want to clarify we dont want this overload 
`  @Nonnull public static Interceptor[] createDefaultInterceptors( @Nullable Interceptor[] interceptors) ` is this accurate ?

baywet: that's my suggestion yes, it doesn't add any value unless I missed something

components/http/okHttp/src/main/java/com/microsoft/kiota/http/KiotaClientFactory.java
raghucha: done

components/http/okHttp/src/main/java/com/microsoft/kiota/http/KiotaClientFactory.java
raghucha: thanks updated the code sorry for confusion

baywet (approved):
I made additional changes to ensure we provide a consistent API surface.

baywet:
@raghucha can you please accept the license agreement? (see the bot message at second comment)

raghucha:

PR #1696, opened 2024-12-09
Handle exceptions when resetting streams after writing to the network

Ndiritu:
We added the `isOneShot()` check to ensure that rewindable request body streams are reset after writing them e.g. after request body is written to a logging interceptor. Previously, a logging interceptor would drain the request body stream causing the network call to have an empty body.

However, we didn't handle the failure scenarios of `reset()` according to the [docs](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#reset--)

This PR makes a best effort measure to prevent draining the request body by an interceptor before the final network request.

If a positive content length is available, we [`mark()`](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#mark-int-) the start of the stream with the entire stream's content length as the read limit.

If during the `sink.writeAll()` a new `mark()` is set causing the `reset()` to fail, we log the exception but don't fail the entire request.

We do not fail the entire request in case the first call to `reset()` happens during the network call.

closes https://github.com/microsoftgraph/msgraph-sdk-java/issues/2245

Existing unit [tests](https://github.com/microsoft/kiota-java/blob/main/components/http/okHttp/src/test/java/com/microsoft/kiota/http/OkHttpRequestAdapterTest.java#L410-L536) for the various stream scenarios still pass

### DIFF
--- a/components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
+++ b/components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
@@ -902,10 +902,8 @@ public class OkHttpRequestAdapter implements com.microsoft.kiota.RequestAdapter 
 
                                 @Override
                                 public void writeTo(@Nonnull BufferedSink sink) throws IOException {
-                                    // stored in variable before writing to the sink due to
-                                    // available()'s definition
                                     long contentLength = contentLength();
-                                    if (contentLength() > 0) {
+                                    if (contentLength > 0) {
                                         requestInfo.content.mark((int) contentLength);
                                     }
                                     sink.writeAll(Okio.source(requestInfo.content));
--- a/components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
+++ b/components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
@@ -904,7 +904,15 @@ public class OkHttpRequestAdapter implements com.microsoft.kiota.RequestAdapter 
                                 public void writeTo(@Nonnull BufferedSink sink) throws IOException {
                                     sink.writeAll(Okio.source(requestInfo.content));
                                     if (!isOneShot()) {
-                                        requestInfo.content.reset();
+                                        try {
+                                            requestInfo.content.reset();
+                                        } catch (Exception ex) {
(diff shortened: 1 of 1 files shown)


ihudedi:
Hi @Ndiritu 
I saw another call to reset in this file and exception is thrown

Thanks,
Itay

ihudedi:
Hi @Ndiritu 
Did you see my last message.In retry method there is also reset that throw an exception.
Thanks,
Itay

Ndiritu:
> Hi @Ndiritu Did you see my last message.In retry method there is also reset that throw an exception.(retryCAEResponseIfRequired) Thanks, Itay

@ihudedi the `reset()` called at this point is necessary since before retrying a request after getting a [claims challenge](https://learn.microsoft.com/en-us/entra/identity-platform/claims-challenge?tabs=dotnet) we need to attempt to reset the body.

In this case, the error handling works as expected by bubbling up the exception to a developer's application.

ihudedi:
Hi @Ndiritu 
Thanks for the response.
Itay

components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
```diff
@@ -902,9 +902,21 @@ public long contentLength() throws IOException {
 
                                 @Override
                                 public void writeTo(@Nonnull BufferedSink sink) throws IOException {
+                                    long contentLength = contentLength();
+                                    if (contentLength > 0) {
+                                        requestInfo.content.mark((int) contentLength);
```
baywet, line 907: why are we marking to the content length? isn't this effectively setting the stream to its end, making it look empty to the client when it comes time to write to the network?

Ndiritu: since we are calling mark() before the stream is read, we are setting the reading position to the beginning of the stream. 
The parameter passed to mark() affects the `reset()` by configuring how many bytes can be read from the stream before the mark is invalid

The parameter passed to `mark()` doesn't affect the start position of the read().

https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#mark-int-

The unit [tests](https://github.com/microsoft/kiota-java/blob/main/components/http/okHttp/src/test/java/com/microsoft/kiota/http/OkHttpRequestAdapterTest.java#L410-L536) confirm this.

baywet (approved):

PR #1775, opened 2025-02-18
Ensures 3XX responses without location header do not throw

### LINKED ISSUE #1132: ensure 3XX responses without location header do not throw
follow up to https://github.com/microsoft/kiota/pull/4367
and https://github.com/microsoft/kiota/issues/4190

Implement a unit test with a 304 response, no location header, check the request adapter returns null and doesn't throw.
Most likely this line will need to be adapted.
https://github.com/microsoft/kiota-java/blob/59a86b031985c34a8f9bff88c9af2ce8eb55c354/components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java#L611

timayabi2020:
Fixes #1132 
- Add checks for 3XX responses without a location header to prevent throwing an exception.

### DIFF
--- a/components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
+++ b/components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
@@ -70,17 +70,24 @@ public class OkHttpRequestAdapter implements com.microsoft.kiota.RequestAdapter 
     }
 
     /**
-     * Instantiates a new OkHttp request adapter with the provided authentication provider.
-     * @param authenticationProvider the authentication provider to use for authenticating requests.
+     * Instantiates a new OkHttp request adapter with the provided authentication
+     * provider.
+     *
+     * @param authenticationProvider the authentication provider to use for
+     *                               authenticating requests.
      */
     public OkHttpRequestAdapter(@Nonnull final AuthenticationProvider authenticationProvider) {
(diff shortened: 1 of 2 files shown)


components/http/okHttp/src/main/java/com/microsoft/kiota/http/OkHttpRequestAdapter.java
```diff
@@ -614,6 +644,15 @@ private Response throwIfFailedResponse(
...
             final ResponseHeaders responseHeaders =
                     HeadersCompatibility.getResponseHeaders(response.headers());
+            if (statusCode >= 300 && statusCode < 400) {
+
+                if (responseHeaders.get("Location") != null) {
+                    spanForAttributes.setAttribute("Location", true);
+                    return response;
+                }
+                spanForAttributes.setAttribute(errorMappingFoundAttributeName, false);
+                return null;
```
andrueastman, line 654: Shoudn't we return the reponse either way to the caller since we aren't throwing an exception?

andrueastman (commented):
Any chance you also align the commit message format for the purposes of release/changelog generation as outlined at https://github.com/microsoft/kiota-java/blob/main/CONTRIBUTING.md#commit-message-format

baywet (approved):

baywet:
@andrueastman for final review and merge (please squash with a conventional commit so we trigger a release here)

PR #1951, opened 2025-06-20
ci/ado esrp migration

### LINKED ISSUE #1890: End-of-Life Sunset Date for OSSRH - June 30th
https://central.sonatype.org/news/20250326_ossrh_sunset/

It is my understanding that this repo is using OSSRH behind the curtains to publish releases.

I think that this is something "owned" by some Microsoft department, but opening an issue for tracking purposes.

baywet:
fixes #1890


scripts/copyPackageContent.ps1
```diff
@@ -0,0 +1,27 @@
...
+
+Write-Output "---------------------------------------------------"
+Write-Output "Copying package contents at $packageFullPath"
+
+if(-not (Test-Path -Path $packageFullPath)) {
+  Write-Output "Package not found in local cache."
+  exit 1
+}
+
+Copy-Item -Path "$packageFullPath\*" -Destination $OutputDirectory
```
peombwa, line 25: @baywet, are we expecting only zip files directly within `$packageFullPath\*`, with no subdirectories present? If there are subdirectories in this path, we should include the `-Recurse` parameter with `Copy-Item` to ensure all contents, including those in subfolders, are copied.

baywet: it's a flat structure, this is due to how the java module packaging works, and I don't believe it'll change anytime soon

PR #1986, opened 2025-07-05
chore: use gradle version catalogs as a single source of truth for dependency

### LINKED ISSUE #1594: Use Gradle version catalogs to have a single source of truth for dependency versions
I propose using Gradle's [version catalogs](https://docs.gradle.org/current/userguide/platforms.html) to have a single source of truth for the project's dependency versions and also (hopefully) reduce the number of dependabot PRs for the same dependency in the project.

exoego:
Closes #1594 

Move all definition of dependencies into [Gradle version catalogs](https://docs.gradle.org/current/userguide/version_catalogs.html).
This will reduce duplicates of same dependency update in multiple sub projects such as:
image

Mostly auto-generated by [the tool I wrote](https://github.com/exoego/gradle-version-catalogs-cli)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,8 +1,8 @@
 plugins {
     id "java-library"
-    id "org.sonarqube" version "6.2.0.5505"
-    id 'com.github.spotbugs' version '6.2.1'
-    id "com.diffplug.spotless" version "7.0.4"
+    alias(libs.plugins.org.sonarqube)
+    alias(libs.plugins.com.github.spotbugs)
+    alias(libs.plugins.com.diffplug.spotless)
 }
 
 sonar {
--- a/components/abstractions/android/build.gradle
+++ b/components/abstractions/android/build.gradle
@@ -8,9 +8,9 @@ buildscript {
     }
 
     dependencies {
-        classpath "com.gradle:gradle-enterprise-gradle-plugin:3.19.2"
-        classpath "com.android.tools.build:gradle:8.11.0"
-        classpath "com.github.ben-manes:gradle-versions-plugin:0.52.0"
+        classpath(libs.com.gradle.gradle.enterprise.gradle.plugin)
+        classpath(libs.com.android.tools.build.gradle)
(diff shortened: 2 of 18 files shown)


baywet:
Hi @exoego 
Thanks for the contribution!

We're receiving multiple pull requests per dependency update because:
- dependencies are used across packages
- dependencies are used both android and the "regular" project

I believe that as long as we don't update the dependabot configuration to contain a single entry, we'll still get multiple PRs per dependency, which will all target the new file you're introducing.

What do you think?

exoego:
> Then, if the intent of this pull request is not to solve the multiple pull requests per dependency, what are you looking to achieve beyond having all the versions at the same place?

It's not MY intention. This PR just solves #1594 which is kept open since 2024 Sep, so I guessed maintainers have wish to implement it.
If you think version catalog is not good to have, then you should close this PR and #1594.

----

Another benefit of version catalog is having version variable in `[versions]` section.
It helps updating some group of relevant dependency at once, which reduce dependabot PRs at some extent.

There is two examples in this PR
- okhttp
- opentelemetry



```toml
[versions]
okhttp = "4.12.0"
opentelemetry = "1.51.0"

[libraries]
...
com-squareup-okhttp3-logging-interceptor = { group = "com.squareup.okhttp3", name = "logging-interceptor", version.ref = "okhttp" }
com-squareup-okhttp3-mockwebserver = { group = "com.squareup.okhttp3", name = "mockwebserver", version.ref = "okhttp" }
com-squareup-okhttp3-okhttp = { group = "com.squareup.okhttp3", name = "okhttp", version.ref = "okhttp" }
...
io-opentelemetry-opentelemetry-api = { group = "io.opentelemetry", name = "opentelemetry-api", version.ref = "opentelemetry" }
io-opentelemetry-opentelemetry-context = { group = "io.opentelemetry", name = "opentelemetry-context", version.ref = "opentelemetry" }
```

peombwa (approved):

PR #2025, opened 2025-09-04
Fix deserialization of `List<Double>` in JSON parse node

jamie-heller:
### Summary
Adds support for `java.lang.Double` in JSON primitive value parsing, fixing failures when deserializing collections of doubles. Includes a unit test validating getCollectionOfPrimitiveValues(Double.class).

### Motivation
Collections of Double values could not be deserialized because JsonParseNode.getPrimitiveValue lacked a `Double.class` branch. This broke scenarios parsing arrays like [1.1, 2.2, 3.3], including on kiota-generated model classes that contain lists of doubles.

### Changes

- Adds handling for `Double.class` in `getPrimitiveValue` by delegating to `getDoubleValue()` in `JsonParseNodeTests`
- Adds getCollectionOfPrimitiveDoubleValues test asserting successful parsing of {"values":[1.1,2.2,3.3]} into a List<Double>.  Prior to the update in getPrimitiveValue, this test failed with `java.lang.RuntimeException: unknown type to deserialize java.lang.Double`

### DIFF
--- a/components/serialization/json/src/main/java/com/microsoft/kiota/serialization/JsonParseNode.java
+++ b/components/serialization/json/src/main/java/com/microsoft/kiota/serialization/JsonParseNode.java
@@ -146,6 +146,8 @@ public class JsonParseNode implements ParseNode {
             return (T) itemNode.getIntegerValue();
         } else if (targetClass == Float.class) {
             return (T) itemNode.getFloatValue();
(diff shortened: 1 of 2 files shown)


jamie-heller:
> Thanks for the contribution!

Thanks for the timely review and release!

PR #2144, opened 2026-07-23
Support map-style query parameters (type: object + additionalProperties)

bkoelman:
Contributes to https://github.com/microsoft/kiota/issues/3800.

Add a Map branch to getSanitizedValues so that plain-object dictionary query parameters are pre-processed before they reach StdUriTemplate.

Without this change, a Map<String, String> value passed through addQueryParameters or addQueryParameter falls through to the else-return branch unchanged.  When the map contains a null value the subsequent call to StdUriTemplate.convertNativeTypes(null) throws a NullPointerException (null.getClass() is called in the fallback throw).

The fix mirrors the kiota-dotnet SanitizeDictionary approach exactly: null-valued entries are dropped (RFC 6570 para 2.3 treats undefined variables as absent), and remaining values are normalised to String via the existing getSanitizedValues pipeline.  The sanitized HashMap<String, String> is returned so StdUriTemplate.isMap/addMapValue can expand it as individual key=value pairs for RFC 6570 {?param*} templates.

The @SuppressWarnings("unchecked") annotation suppresses the unavoidable raw-type cast of Map<Object,Object> -- Java generics are erased at runtime so the cast to the generic Map type is unverifiable by the compiler, but safe because the entries are accessed only as Object.

### DIFF
--- a/components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
@@ -482,12 +482,10 @@ public class RequestInformation {
             // the variable expansion results in no output". Use "" to send ?key= (empty value).
             @SuppressWarnings("unchecked")
             final Map<Object, Object> rawMap = (Map<Object, Object>) value;
-            final HashMap<String, String> sanitized = new HashMap<>(rawMap.size());
+            final HashMap<String, Object> sanitized = new HashMap<>(rawMap.size());
             for (final Map.Entry<Object, Object> entry : rawMap.entrySet()) {
-                if (entry.getValue() != null) {
-                    sanitized.put(
-                            entry.getKey().toString(),
-                            getSanitizedValues(entry.getValue()).toString());
+                if (entry.getKey() != null && entry.getValue() != null) {
+                    sanitized.put(entry.getKey().toString(), getSanitizedValues(entry.getValue()));
                 }
             }
             return sanitized;
--- a/components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
+++ b/components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
@@ -476,6 +476,21 @@ public class RequestInformation {
                 return result;
             }
             return Arrays.asList(values);
+        } else if (value instanceof Map) {
+            // Pre-process map/dictionary values for StdUriTemplate expansion.
+            // Null values are skipped: RFC 6570 §2.3 states "if the value is undefined,
+            // the variable expansion results in no output". Use "" to send ?key= (empty value).
+            @SuppressWarnings("unchecked")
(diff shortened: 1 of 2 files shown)


components/abstractions/src/main/java/com/microsoft/kiota/RequestInformation.java
```diff
@@ -476,6 +476,21 @@ private static Object getSanitizedValues(Object value) {
                 return result;
             }
             return Arrays.asList(values);
+        } else if (value instanceof Map) {
+            // Pre-process map/dictionary values for StdUriTemplate expansion.
+            // Null values are skipped: RFC 6570 §2.3 states "if the value is undefined,
+            // the variable expansion results in no output". Use "" to send ?key= (empty value).
+            @SuppressWarnings("unchecked")
+            final Map<Object, Object> rawMap = (Map<Object, Object>) value;
+            final HashMap<String, String> sanitized = new HashMap<>(rawMap.size());
```
baywet: same as dotnet, can we use map string object instead to avoid the call to tostring below?

baywet (changes requested):
Thanks for the contribution!

baywet:
CC @andreaTP

components/abstractions/src/test/java/com/microsoft/kiota/RequestInformationTest.java
```diff
@@ -473,6 +473,149 @@ void ReplacesEnumValuesPathParameters() throws IllegalStateException, URISyntaxE
...
+
+        requestInfo.addQueryParameter("%24top", 5);
+        final Map<String, String> map = new HashMap<>();
+        map.put("filter", "active");
+        requestInfo.addQueryParameter("query", map);
+
+        // Assert
+        final URI uri = requestInfo.getUri();
+        final String uriStr = uri.toString();
+        assertTrue(uriStr.contains("%24top=5"));
```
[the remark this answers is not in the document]
baywet: nope, we want to ensure the literal is encoded as expected to avoid breakage of the std uri template lib

andreaTP:
seems like a legit change to me :+1:

bkoelman:
Pushed fixes

components/abstractions/src/test/java/com/microsoft/kiota/RequestInformationTest.java
```diff
@@ -665,7 +693,7 @@ class GetQueryParameters implements QueryParameters {
      * Free-form key-value query parameters (type: object + additionalProperties).
      * Expanded via RFC 6570 {?query*} into individual key=value pairs.
      */
-    @jakarta.annotation.Nullable public Map<String, String> query;
+    @jakarta.annotation.Nullable public Map<String, Object> query;
```
baywet, line 696: can we also add tests to validate that map string, string | int | float, etc... (primitive types essentially) also normalize as expected please? 

(as I clarified in my similar comment in dotnet, I would expect the "correct type" to be generated, but the internal methods to still use object)

